PR #4, opened 2016-05-16
MPI support

csullivan:
This adds support for looping over ATTPC root trees with the MPIEventLoop library. 

I noticed that AtTpcMap.h seems to be missing in the develop branch. This prevents me from testing the merge locally, so I would suggest checking out this pull request, resolving the header issue, and then testing to see if all compiles and runs. 

FYI: The MPI branch in my repository is fully functional, but is behind the develop branch.


csullivan:
@Yassid I copied the missing header over and pushed the commit (1db55914f7796cbfdeda340b37bfd23a6f1c4a78) directly to the develop branch to avoid another pull request. Everything works now as expected. Cheers

PR #11, opened 2017-07-25
Done: exporters

freund17:
fixed pcl exporter
added json exporter

I was told an exporter with fileFormats readable by python could be useful. So I started one.


Yassid:
HowTo is needed for these scripts/macros.

Issue #13, opened 2017-07-26
More trajectory info when using hierarchical clustering


freund17:
Currently the trajectories are just a vector of hit-indices.
In the near future this will be replaced by a new class **ATTrajectory**, which will contain the following information:
  - indices of points belonging to that trajectory
  - average curvature
  - start index (index of the point at the very start of the trajectory)
  - end index (because, why not?)
  - length of the trajectory (following the curve; using a
minimum-spanning-tree and shortest path from start to end)
  - centerpoint of the trajectory (average position of all points belonging to that trajectory)
  - main direction (basicly a PCA)

Yassid:
I suggest to modify the existing ATTrack class, for better compatibility with other parts such as the MC fit.

PR #14, opened 2017-07-26
WIP: issue #13

freund17:
Not jet ready to be merged, but you can take a first look at how things are coming together.
Feel free to give the code a testrun using the provided test macro. (`macro/examples/run_hierarchical_clustering.C`)

For the moment the new class `ATTrajectory` is used (making `ATHierarchicalClusteringCluster` obsolete).
For integration into `ATTrack` I would need more information about its properties, where it is used and how.

@Yassid I also rendered some of the lines using `GetCentroidPoint` and `GetMainDirection` as described in the email we exchanged.


freund17:
`GetApproximateTrajectoryLength` still needs to be implemented.

Yassid:
Very nice. ATTrack is essentially used in the RANSAC, Hough and Analysis modules. It contains the associated Hit Array (std::vector<ATHit>) and geometric and simulated/fit properties of the track. By geometric I mean, parameters extracted from the pattern recognition algorithm. You parameters could be classified as geometric because no fit is applied. I noticed you use Eigen vectors. ATTrack uses TVector3. Maybe the easiest way to integrate this is make ATTrajectory a child class. 

It would be nice to know the run and event number of the tracks/trajectories you show.

freund17:
I'm not happy with the way the average curvature is calculated...
Any ideas?

Yassid:
Few things are possible:

1.- Circle fit to the first part of the track that has a better defined circular shape (circular fit).

2.- Use the Hough transform for circles in a very reduced space. You have an example here: https://github.com/ATTPC/ATTPCROOTv2/blob/develop/reco/HoughTransformation/ATHoughSpaceCircle.cc#L103

3.- Use this geometrical calculation (see figure)
 in a histogram to find where the lines converge. Instead of doing it for each neighbor point you do it each ten points, for example.

screen shot 2017-07-27 at 10 05 12 am

freund17:
Just a few more screenshots also showing the `GetApproximateTrajectoryLength`. (for documantary reasons)
line-length
lineandcurve-length
lineandcurve-length2

Yassid:
Just a comment: When working under magnetic field, we don't use the range to extract the energy of the particle but the magnetic rigidity (for the curved trajectories obviously), which is the magnetic field times the curvature. In such a case, the range is not really very important. In any case, it would be nice to have some approximated value.

freund17:
ATCubicSplineFit is now working as expected.
We can now calculate as much points on a smooth curve as we want.
smoothinterpolatedlinefixed

I also added the functionality to skip points:
smoothinterpolatedlinefixedjump2

But smoothening the point-data beforehand will probably be more promising.

freund17:
I tested ATCubicSplineFit on a smoothed cloud:
smoothinterpolatedlinesmooth
The results are better than before, but there are still three main problems:
  1. The generated points are not spread out evenly
  2. The points are not necessarily in the right order
  3. Some points may be projected on the same point, making differentiation impossible

Using the skip functionality you can work around some of the issues:
smoothinterpolatedlinesmoothjump

But skipping doesn't solve all of the issues and it introduces jet another parameter, wich likely changes for every trajectory.

What I'm trying to do now is to project every point onto the `mainDirection` and use it's position on there as the `position`-parameter in the `ATCubicSplineFit` algorithm, which should solve all of the above issues in an elegant way.

freund17:
Now the position-projection I mentioned above is implemented. It guarantees to generate evenly distributed points and serves as a metric to sort them.
smoothinterpolatedlinesmoothprojected
smoothinterpolatedlinesmoothprojected2

Using this, we have a good distance metric between points and are able to skip them if they are to close together. That way we can create an even smoother curve.
smoothinterpolatedlinesmoothprojectedmincontroldistance

freund17:
Curvature calculation still wrong...
I'll try this tomorrow:
https://en.wikipedia.org/wiki/Curvature#Curvature_of_space_curves

freund17:
As far as I can tell, everything seems to be working now...
@Yassid Mabe you can check the results of the curvature caluclation, as I'm still not sure about it.
I identified one of the problems of my previous approches as running into accuracy issues with the `float` datatype.

Yassid:
The best way to check if the curvature calculation works is to compare with the Hough Space circle calculation. Once the code is ready, we can test it on 46Ar data. Actually, the comparison will be a perfect benchmark for the paper.

Yassid:
Other way is to prepare test data where we have simple circles projected onto the XY plane.

freund17:
Attention: Points which are not detected as part of a trajectory are no longer visualized.

freund17:
It is now once again possible to display points which do not belong to any trajectories.

freund17:
I did some parameter-tweaking, which fixed most of the problems you mentioned in your e-mail.
Trajectories separating at a small angle are still a problem.

flatangle

freund17:
A more agressive `GenTripletsMaxError` fixes the problem in this case.

flatanglemoreagressivegentripletsmaxerror

freund17:
I'm leaving today and did what I could. @clsantama will take it from here.
The average curvature is still a problem.
With the code I added, the vertex should easily be found. The shape of the trajectories shouldn't matter. The end of trajectories will be linearly extrapolated when necessary. But be told that it expects every passed Trajectory to cross in a single point (more or less).
So something like this:

greatcrossing

Would work great, whereas something like this:

badcrossing

would yield wrong results. An easy fix would be to only take the two best trajectories into account.
(Btw.: the trajectories in these screenshots are extrapolated by ~100 units at both ends)

About the simulations:
There seem to be some scaling issues, which lead to my algorithm merging a lot of trajectories together. This could be fixed by tweaking the parameters, but I found it to be much more easy to simply scale-up the input data by a factor of 10.

Yassid:
Just one comment, can the vertex extrapolation be disabled easily to do tests? As first approach I would like to see if the MC can reconstruct the tracks by giving the first point of the trajectory the code finds.

The simulation is being fixed, I fixed several problems about the kinematics.

freund17:
When you have a look at the example, you will see that the search for the vertex is just a simple function which is applied after "my algorithm" has found the trajectories. You basically pass it a `vector` of `ATTrajectorie`s you want to find the collision-point for.
For the extrapolation part, this is automatically done in `ATCubicSplineFit` whenever you request a position which is outside the boundaries of `cubicSplineFit.GetStartPosition()` and `cubicSplineFit.GetEndPosition()`.

Yassid:
Awesome, is this ready to be merged then?

freund17:
According to me...
It is your code now.

PR #15, opened 2017-07-27
WIP: refactoring ATHoughSpaceCircle

freund17:
I started refactoring `ATHoughSpaceCircle`.
Currently I only started on `void CalcHoughSpace(ATEvent *event, TH2Poly *hPadPlane)`.
I
  - added spacing
  - corrected indentation
  - used vectors and vector arithmetic when applicable
  - replaced `TMath::Power(..., 2)` with explicit `(...) * (...)` (in some cases it can be faster that way)
  - changed `i++` to `++i`
  - replaced `for (index = 0...)` with `for (... : ...)` when applicable
  - began moving isolated parts of code in their own functions
  - added `const` to all getters
  - commented and marked (with `TODO`) unused code


freund17:
I might add, that I could not jet test it. But it compiles without warnings.

reco/HoughTransformation/ATHoughSpaceCircle.cc
```diff
@@ -90,338 +90,352 @@ TH2F* ATHoughSpaceCircle::GetHoughSpace(TString ProjPlane)
...
+
+        if (geo_dist < 10.0)
+        {
+            IniTSBuff = hit.GetTimeStamp();
+            IniRadiusBuff = fRadius->at(iHit);
+            IniPhiBuff = fPhi->at(iHit);
+
+            if (hit_dist < 10.0)
+            {
+                fClusteredHits->push_back(&hit); // TODO: why pointer? pointer to stack? why not copy?
```
freund17: Why is a pointer passed here?
In the original version `hit` was a copy of the element in `event->GetHitArray()->at(iHit);`
```c++		
ATHit hit = event->GetHitArray()->at(iHit);
```
This copy would have been on the stack, meaning it would get deleted as soon as the function is left and overwritten as soon as the next loop comes around.
I changed it to be a reference to the element in the original array instead.
```c++
 ATHit &hit = event->GetHitArray()->at(iHit);
```
But I would prefer to not touch `event` here and to make it `const`.
```c++
 ATHit const &hit = event->GetHitArray()->at(iHit);
```
But that would mean `fClusteredHits` would need to hold a `const` pointer or a copy.
A `const` pointer would make more sense, if we can guarante that the event will not be deleted before the `ATHoughSpaceCircle`. Otherwise a copy would be needed here.

Yassid: Indeed, make a copy, I don't even know why it is passed as a pointer.

FYI: Clustered hits contains one "mean hit" per time bucket (time unit).

Yassid: By the way, regarding pointers, we should start using unique and shared ptr either from std or boost.

clsantama: We can use boost, as we short-listed it in the list of needed libraries for the final C++ classes.

freund17: Regarding pointers, we should stop using new and delete entirely.
They only belong in data-structure-classes (like vector, list, etc.)

And you do not need to pass a pointer every time a big object is returned. I'm a huge fan of `const &` if you didn't see already. And since move-constructors are a thing, a normal "copy" of a vector, for example, isn't a problem either.
And then, what you are left with, are most of the time simple pointers which just help you find that element in a array of your own object. And that kind of pointer doesn't need to be managed.

freund17:
@Yassid What about this unused code?: https://github.com/ATTPC/ATTPCROOTv2/blob/c736f81a46b8d68cd943c69875a4d576bdc62ed8/reco/HoughTransformation/ATHoughSpaceCircle.cc#L201

Yassid:
If thats old code that was used in the previous way we calculate the radius of curvature, it can be removed.

PR #16, opened 2017-07-31
Calibration

watwoodn:
Added a calibration class to ATPulseAnalyzer. Also a calibration folder to Unpack_GETDecoder2 with sample calibration files and macro used to generate these files from a pulser file


Yassid:
The txt files containing the calibration data should be stored with better names under dedicated folders, i.e.: full-scale or prototype backgammon pad plane, measurement with/without THGEM and all these thing that will avoid confusion in the future.

PR #21, opened 2017-09-11
removed cloudScale calculation

freund17:
removed cloudScale calculation to have codebase in sync with dalitz' version


freund17:
As I don't have the Root-Framework installed, I was unable to test this.
Please test, before you merge.

PR #28, opened 2018-12-13
few corrections

jczamorac:


Yassid:
The HDF parser has to be compliant with the new data format. You should implement some flag for old formats (like the 22Mg experiment) to ensure the compatibility.

PR #42, opened 2020-08-24
offline merging and analysis S800-ATTPC

giraudsimon:
-added libraries for s800-ATTPC offline merging and analysis.
-updated beam shape simulation

! WARNING ! Maybe an issue in ATRansac. 
(please see the class ATRansac, function CalcRANSAC),
without the comment line std::cout<<"RansacPCL tracks size : "<<tracksSize<<std::endl; the fit is not done (code runs but not fit), on flagtail. It was already the case before implementing the new stuff.


Yassid:
I will accept it but someone will have to debug the RANSAC problem, which is extremely weird.

PR #44, opened 2020-08-27
basic ransac class

jczamorac:
Basic RANSAC class.
It was implemented but not tested. Probably a few changes have to be included before using. I will need help to include the visualization


Yassid:
I am accepting the pull request but this class will be moved into ATPatternRecognition

PR #45, opened 2020-09-02
Fixes from e12014 experiment

anthoak13:
Two changes are allowing the full range of time bucket options and recording the timestamp into in ATRawEvent, I missed that in the first pass.,


Yassid:
Actually, important for the next experiment

PR #46, opened 2020-09-02
Fix a stupid mistake in the merge

anthoak13:
I accidently left two version of the time bucket switch statement in the code. I also recently found Alex Carls thesis that has a decent overview of the framework, so I added a link to that in the readme.


Yassid:
We should actually complete the manual that Alex wrote.

PR #48, opened 2020-09-04
Added working fission generator and made gas fano factor and GET electronic gain modifiable parameters in ATDigiPar file

anthoak13:


anthoak13:
Realized I forgot to add the shaping time to the digipar file too. Fixes that then will resubmit pull requewst

PR #54, opened 2020-09-22
S800-ATTPC work

jczamorac:
* A flag for S800 gate was included into TRaw and TEvent
* A new folder for the experiment was created
* TCutG reader was modified to include several cuts with the same function


Yassid:
This is a temporary modification. Gate management will be moved to the Analysis task.

PR #56, opened 2020-09-25
Add inverse lookup to AtTpcMap

anthoak13:
I did this the simplest way possible sot here are two copies of the map, but I though the increase in memory was worth not having to over-complicate things by using pointers to reuse memory.


Yassid:
Just to be sure since the file comparison format is a little bit messed up. Is this just an extra member function for the map class?

anthoak13:
I added another member function and variable. The variable is map<int, vector<int>>, the member function uses that member to do the reverse lookup.

Yassid:
Got it thanks.

PR #57, opened 2020-10-26
FFT first version

DanielRegueira:
FFT first version Dani


Yassid:
Some comments before accepting the request:
1.- Please, do a git pull to accept the new changes in the development branch (save your work before).
2.- Rename macros to avoid overwriting the previous ones (e.g. ATTPC_Mg22elastic.C).
3.- pdfs are not allowed in the repository. Please, add them to the git ignore list or remove them.
4.- Avoid personal names in the folder description (e.g. Hierar_Dani). From now, we will move all the existing macros under the /macro folder (for example /macro/e15250). You can use a more appropriate definition inside this folder. This is to be consistent with where your simulation macros are, otherwise it becomes very confusing.
5.- Whats the reason for chaging the HDFParser so dramatically?

Yassid:
Waiting for an updated version of the FFT

PR #74, opened 2021-02-23
Large refactoring of code. Will break macros, but not badly.

anthoak13:
A list of changes:

1) All of the source code was refactored so the file extensions are consistent throughout the code base and agree with the ROOT standard. Source files are cxx and header are h.

2) The directory structure was changed so it is consistent. Each folder that begins with "At..." is built to a library with the name "libAt...". This primarily effects the folders reco, digi, par, and s800 which for some reason deviated from everything else. Folders that begin with lower case letters like macro and parameters were left unchanged as they are not built into a library and are not part of the CMake build process.

3) The names of classes were standardized. Throughout the code base, the prefix on classes was sometimes "AT" and other times "At". All of there were changed to "At" to be consistent with the naming standards used by ROOT. This sometimes swept up other parts of class names. For example `ATPSATask` became `AtPSAtask`.

4) All of the code was reformatted to be consistent with ROOT's code style. All future additions to the code should maintain the same style. This will fix the problem of it being very hard to tell what was changed in a pull request when the code was reformatted. There is a `.clang-format` file provided (from ROOT Dev team) that will enforce the code style when the command `clang-format -i [files]` is ran. Running git-clang-format is a shortcut to only format the lines of code that were changed in since the last commit. Vim and Emacs both have support for code formatting using clang-format. Google it.

5) The readme was updates with information for developers on the coding style. It was taken from the ROOT project which this code is derived.

6) The pulse shape analysis part of the code was refactored. Details on this were added to the readme. Basically, any changes to the public method of one of the derived PSA methods meant that the base class `AtPSA` and the task class `AtPSAtask` also had to change. This is not good. Rather then methods being passed from the task to the class actually handling the logic, they can be called directly on the logic class themselves.

I tested things as best I could, and I think everything is working. Because library names changed, you will need to do a distclean or delete the build directory and rebuild everything. If something seems badly broken, let me know. I didn't touch the macro directory so you will have to update your macros individually. It should be a simple matter of find-replace "AT" with "At", avoiding replacing anything like an environment variable. A working example of a HDF5 unpacking class was added to the folder `macro/examples`.


anthoak13:
I think I might have found a bug in the PSA part of the code I changed. I want to try to track it down before this gets pulled in.

anthoak13:
This commit fixes the bug. I forgot that I had to move anything that reads the parameter file to a method that only gets called after the parameter file is read (so somewhere in the Init stage).

Yassid:
Thanks, I will review this as soon as I finished refactoring the map class (almost done)

anthoak13:
No rush. I just wanted to make sure this wasn't merged when I knew there was a bad bug in it.

Yassid:
I am accepting the pull request, we need to incorporate the new mapping recently pushed to master and develop_prerefactoring branches. Essentially, the files to merge are the entire AtTpcMap and eventDisplayAT folders, and the ATCore.cc, ATCore2.cc and ATHDF5UnpackTask.cc. They have been refactored to take the new map hierarchy into account

Yassid:
To be more specific

AtMap/AtTpcMap.cxx: needs merge
AtMap/AtTpcMap.h: needs merge
AtMap/AtTpcMapLinkDef.h: needs merge
AtMap/AtTpcProtoMap.cxx: needs merge
AtMap/AtTpcProtoMap.h: needs merge
CMakeLists.txt: needs merge
eventdisplayAT/ATEventDrawTask.cc: needs merge
eventdisplayAT/ATEventDrawTask.hh: needs merge
eventdisplayAT/ATEventDrawTaskProto.cc: needs merge
eventdisplayAT/ATEventDrawTaskProto.hh: needs merge
eventdisplayAT/ATEventDrawTaskS800.cc: needs merge
eventdisplayAT/ATEventDrawTaskS800.hh: needs merge
eventdisplayAT/ATEventProtoDrawTask.cc: needs merge
reco/ATDecoder/ATCore.cc: needs merge
reco/ATDecoder/ATCore.hh: needs merge
reco/ATDecoder/ATCore2.cc: needs merge
reco/ATDecoder/ATCore2.hh: needs merge
reco/ATHDFParserTask.cc: needs merge
reco/ATHDFParserTask.hh: needs merge

PR #77, opened 2021-03-26
Merge in refactored map class and Genfit2

anthoak13:
I wasn't able to test the Genfit2 part, so almost certainly there will be build errors that need to be fixed. It should just be renaming things to match. CMake didn't find genfit on fishtank and trying to get both the code to link with an additional library while also refactoring seemed like a recipe for disaster.

Because of the lack of genfit on my end, I commented out the lines that required it on the AtReconstruction LinkDef and moved things around in the CMakeList.txt so everything that requires genfit is tagged as such.


anthoak13:
@Yassid A question I had working on this, why are there what seems to be three different ways of drawing prototype events? There is an option in AtDrawTask for the prototype [here](https://github.com/ATTPC/ATTPCROOTv2/blob/706eecaf7ee7199f7ae28e6a4f6c43e94855bd9a/AtEventDisplay/AtEventDrawTask.cxx#L134) . There are also two classes, AtEventDrawTaskProto and AtEventProtoDrawTask, the latter inherits from AtEventDrawTask the former has its own code. We should think about simplifying this somehow.

anthoak13:
> @Yassid A question I had working on this, why are there what seems to be three different ways of drawing prototype events? There is an option in AtDrawTask for the prototype [here](https://github.com/ATTPC/ATTPCROOTv2/blob/706eecaf7ee7199f7ae28e6a4f6c43e94855bd9a/AtEventDisplay/AtEventDrawTask.cxx#L134) . There are also two classes, AtEventDrawTaskProto and AtEventProtoDrawTask, the latter inherits from AtEventDrawTask the former has its own code. We should think about simplifying this somehow.

I think, if I understand dynamic_cast correctly, that with the map refactor if someone were to try to use the prototype option in `AtDrawTask` they would get a null-pointer exception. When the prototype option is passed an instance of `AtTpcProtoMap` is created and registered with ROOT as ["fMap"](https://github.com/anthoak13/ATTPCROOTv2/blob/698caf42a8e82f77eabb5fae6e8a96bef2b39450/AtEventDisplay/AtEventDrawTask.cxx#L145) as type `AtMap`. It's then loaded from the root globals [later](https://github.com/anthoak13/ATTPCROOTv2/blob/698caf42a8e82f77eabb5fae6e8a96bef2b39450/AtEventDisplay/AtEventDrawTask.cxx#L1692) and dynamic cast to `AtTpcMap` which will return a null pointer if the most derived type of "fMap" is `AtTpcProtoMap`. Probably that option should just be removed, and then the member variable could be changed to AtTpcMap and the dynamic casts would be unnecessary.

Yassid:
@anthoak13 Thank you very much, I will take care of the merging from now,  I need to finish the GENFIT integration ASAP. 

Regarding the eventDisplay, the one that inherits from the main class (ProtoTask) is totally deprecated and it should be removed. As you can see the new version is used in the latest analysis we did with the prototype (i.e. https://github.com/ATTPC/ATTPCROOTv2/blob/develop/macro/Unpack_HDF5/S1845/run_eve_proto_lite.C). So anything related to the prototype in the main class can be removed (feel free) and leave the stand alone class functional. There a other things to be cleaned too.

The dynamic_cast is a little bit more subtle. The AtMap class was very poorly designed in the sense that everything was going to be inherit from the main class. This leaves no room for a proper hierarchy and even after refactoring I had to patch the code with casts. What I tried to do is to leave a virtual class from where everything can be inherit and adapted (same as you see for example in the PRA class where everything is properly done). In this case, the dynamic cast is needed to let the compiler know that only the derived class has the function we want to use (i.e. SetProtoMap used in Core is unique to the prototype so there is no need to declare it as virtual in the main class).

The functionality of the Core and Unpacker was tested for every detector and it is working. Feel free to modify it if you feel.

anthoak13:
I left everything with the Core, Unpacker, and S800 viewing tasks alone because I'm less familiar with that part of the code, but I cleaned up the deprecated prototype viewing classes (the only valid option is now to use `AtEventDrawTaskProto`. Because each detector has it's own draw task, I removed the now unnecessary dynamic_casts in the draw task and set the pointer to just be of the right derived class.

I'm going to merge this pull request in. We can revert it if we need to, but I'd like to get some of the bugfixes in for Chenyang because he's working off a fork of this branch in the main repository.

PR #79, opened 2021-04-28
Add task to filter traces

anthoak13:


Yassid:
@anthoak13  Just for reference, please comment on the filter purpose.

anthoak13:
@Yassid All this filter does is divide the signal size in half, so not any use. This is just adding in the task and interface to allow others (Chenyang in this case) to easily add filters into the analysis pipeline without having to muck around with data management. I'll put something on the wiki describing exactly what it does but tldr there is a task that is created with a corresponding filter (which must derive from the interface AtFilter). AtFilter requires three methods are defined, and init method and two methods that are passed the raw/baseline subtracted pad traces to modify. The task then calls this filter on every pad in an event and saves the new event on the cbmsim tree with the branch name "AtRawEventFiltered"

anthoak13:
Wiki documentation: https://github.com/ATTPC/ATTPCROOTv2/wiki/Filtering-pad-signals

PR #84, opened 2021-05-11
Added much more detailed detector geometry.

Nuclear-T-Wheeler:
Added more components to detector geometry (end caps, cathode, micromegas, internal support structures, etc.).


anthoak13:
Please reformat the code using clang-format so it matches the coding standards, there's documentation on the wiki for doing this https://github.com/ATTPC/ATTPCROOTv2/wiki/Coding-conventions  then resubmit. Thanks!

PR #85, opened 2021-05-12
Action to check formatting on commit or PR

anthoak13:
Will not check the macro, script, or compiled directories. I left those out because most of the stuff there is individual and there isn't a need to impose uniformity on those bits. Individual groups/experiments can do that if they want. It's easy to add other folder to the exclude list in the workflows/main.yml file if needed

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,19 @@
+# Run clang-format
+name: Clang-format
+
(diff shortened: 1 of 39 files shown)


anthoak13:
Tyler noticed that fishtank only has clang-format version 11 installed. This uses version 12 so I'll change it over to 11 too for consistency then resubmit.

PR #86, opened 2021-05-12
Tyler fork

Nuclear-T-Wheeler:
Added more components to detector geometry (end caps, cathode, micromegas, etc.). Ran git clang-format command on staged file.

### DIFF
--- a/geometry/GADGET_II.C
+++ b/geometry/GADGET_II.C
@@ -229,6 +229,5 @@ TGeoVolume* create_detector()
   gGeoMan->GetVolume(geoVersion)->AddNode(cathode_mount,1,new TGeoCombiTrans(0.0,0,cathode_mount_length/2,new TGeoRotation("cathode_mount",0,tpc_rot,0)));
   downstream_cap->SetTransparency(80);
 
-
   return drift_volume;
 }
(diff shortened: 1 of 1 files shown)


anthoak13:
It looks like it didn't actually run modify the file. It should look something like this (before your edits) https://github.com/ATTPC/ATTPCROOTv2/blob/946b5c934e73caed204ded51df5ed250908007a4/geometry/GADGET_II.C

What command did you run/did it say it modified that file? If you ran `git-clang-format` it will unstage the file, so you have to re-add it.

Nuclear-T-Wheeler:
Hello,

Now every time I try the command it just says that there are no modified files to format:
[cid:86d133fa-3abe-42f4-8c20-51656266533f]

I've never used this before so I'm not sure why it's not working. Any help you can provide would be greatly appreciated. Thanks for your time!

Best,
Tyler
________________________________

anthoak13:
Try `clang-format -i fileName`. The `-i` flag tells the program to modify the file directly, otherwise it just dumps the modified file to stdout. The git-clang-format is basically a wrapper that will run that command on every file that's been staged, but since it was committed I'm guessing you don't have any staged files anymore. 

I'll put a not about this on the wiki. It isn't super clear what to do if the files changed have already been committed.

Nuclear-T-Wheeler:
Thanks so much for your help!

Now, I'm getting this error:
[cid:3b362e03-d6ba-46a0-92b7-01c88344a986]
________________________________

Yassid:
If no objections and as stated by a parallel commit, I will accept this file as it is since it is not a "core" file.

anthoak13:
Sounds good to me. I'll format it as part of the GitHub action to check formatting.


On May 13, 2021 05:31:44 Yassid ***@***.***> wrote:

If no objections and as stated by a parallel commit, I will accept this file as it is since it is not a "core" file.

PR #87, opened 2021-05-13
Github action to check formatting on any PR or commit

anthoak13:
Won't check the macro, scripts or compiled folders. Also only checks .cxx, .h, and .C files. Uses clang-format version 11 for the check.


anthoak13:
@Yassid I know it's possible to add an action that will reformat any files pushed automatically and add the changes as a new commit, do you want me to look into doing that?

PR #91, opened 2021-07-12
Proton and alpha channel added

Ruchi-GADGETII:
Added a proton of 1.2MeV , alpha particle of 0.506 MeV and a gamma ray of 4.03 MeV. 
Added branching ratios for above particles.


Ruchi-GADGETII:
To fix the mirroring of the events, a new parameter "fvariable" is defined which can be changed for GADGETII and default case.
fvariable =kFALSE for default cases and fvariable = kTRUE for GADGETII

PR #96, opened 2021-09-21
NEW PULL REQUEST

Ruchi-GADGETII:
To fix the mirroring of the events, a new parameter "fvariable" is defined which can be changed for GADGETII and default case.
fvariable =kFALSE for default cases and fvariable = kTRUE for GADGETII


anthoak13 (changes requested):
Please run clang-format and amend the commit so it passes the checks before its pulled in. And can you also give fvariable a more descriptive name. fdet_length, as used, should also be an local variable in AtPSA::CalcualteZGeo I think. It is constrained to always be 400 as written. 

I think you also have a bug in the code. fvariable should probably not be hard coded to false, but there should be someway to set the value.  Otherwise the GADGET specific code will never run unless that variable assignment is changed, and the code recompiled. As written write now, the compiler will likely produce identical machine code to the original function.

Since you have to amend the commit anyway, it would be good to make the commit message more descriptive. Your comment on the pull request won't appear on the commit log, just the name of the pull request.

Ruchi-GADGETII:
I have made all the requested changes.
This variable is specially defined for GADGETII detector in order the correct the events that are happening on the opposite side of the detector. For this a new variable fswitch is defined. fswitch = kFALSE for default cases and for GADGETII fswitch = kTRUE

Yassid:
Merge conflict format has to be removed.

anthoak13:
It looks like clang-format still has to be run on some files to fix the formatting. You can call it using `clang-format -i fileName`. The -i flag tells the program to modify the file in place.

I think this is ultimately up to @Yassid but I'm very uncomfortable hard coding a flag for detector type in a method like this. It requires anyone using GADGET to remember this flag exists and they have to change it and rebuild. It also means that any build of the code can be used either for GADGET or for the other detectors. Switching detectors would require a recompile. There were two ways that occurred to me that would help get around this issue:

1. (my favorite) Mark the function virtual, and extend whatever PSA method is being used for GADGET and override this function.  This has the benefit that you can't forget to set the flag (there is no flag anymore), and if more things need to be changed for PSA with GADGET in the future there is already a logical place to do that work without cluttering the base class with if statements. 
2. Make these member variables (like you initially had) but do not set their values in the function. Instead, set defaults in the constructor and provide getters/setters for the variables so the behavior of the class can be changed without recompiling. This has the downside that it clutters the base PSA class unnecessarily.

Yassid:
To be fair I didn't read it in detail so I thought there was a method to select the proper calibration somehow, with member variables. Ideally I would adopt the virtual function solution, but this requires some effort. We can initially adopt the easdy solution (2) so they can move forward with the analysis and then change accordingly.

Yassid:
Looks good to me, @anthoak13, please review and approve.

PR #118, opened 2022-05-25
Cleanup AtPSA, AtHit, and AtDigiPar

anthoak13:
General cleanup:
* Move trackfinder to subdirectory and remove most clang-tidy checks from this external code
* Remove unused files that depend on PCL
* Remove unused parameters from AtDigiPar, old files with extra parameters will still continue to work.

Cleanup of AtPulseAnalyzer:
* Deprecated AtPSASimple2 and split into two classes AtPSAMax and AtPSASpectrum. AtPSASimple2 was
really two distinct PSA methods hiding in the same class.
* Remove more non-owning raw pointers from AtPSATask. Switch to using
a fixed size TClonesArray and just clearing it instead of deleting and re-initializing it.
* Remove AtPSAProto, it was replaced by AtPSASimple2 and then AtPSAMax and AtPSASpectrum.
* Rename AtPSAProtoFull to be more descriptive. The PSA method had
nothing to do with the prototype detector. Also moved the few
variables in the base class that were just used by this method to the method.
* Add Clone() method to AtPSA (so we can copy PSA methods)
* Remove unused parameters from AtHit
   - PositionCorr: Used for tilting the detector
   - IsAux: Unsed by rest of code
   - SlopeCnt: Unused by rest of code
   - BaseCorr: AtPSA should not be modifying traces, should use AtFilter instead (and it is not read anywhere by the rest of the code)
* Refactor AtPSAMax and deprecate old AtPSAtask constructor to remove another reference to non-owning pointers


anthoak13:
It shouldn't affect the macros at all (except for printing a warning about a class being deprecated). I switched this back to draft because I realized there was one additional thing I wanted to add for clustering that would make sense to be in this commit. 

This is a bit of a wider point, but I think it's really important now that many people are starting to work off this same branch that we keep the public interface of the code stable. So deprecate instead of remove/rename. And then only break the public interface on major version updates. This is why I recommended we switch to using semantic versioning which makes how the public interface changes very explicit (and makes it easy to tell from version numbers when there might be a conflict).

anthoak13:
The one exception to what I just said in this commit is the rename of AtPSAProtoFull to AtPSAHitPerTB, but I don't think anyone is currently using that PSA method...

anthoak13:
I tested it, and it looks like these changes will still work with old data (the changes to AtHit don't require unpacking again as far as I can tell) so I'm going to merge it in.

PR #131, opened 2022-07-11
Add additional options for beam simulation.

cnhunt:
Add option to disable reaction in simulation. Add method to set beam
emittance in simulation.


anthoak13:
Closing because of formatting.

Issue #134, opened 2022-07-12
Weighted Y sample bug


anthoak13:
Weighted Y sample will hang at certain events. Probably when all the hits in the event are vetoed.

PR #135, opened 2022-07-12
Fix WeightedY sampling hanging (#134)

anthoak13:
Reverts to just charge weighted when every hit is inside or outside
the beam region


anthoak13:
Fixes #134

PR #145, opened 2022-08-04
Refactor AtHitCluster, AtEvent, and AtPSA

anthoak13:
Adds to AtHitCluster so it can track it's own mean and variance as hits are added. (Can still manually set the position and cov)

Refactor AtEvent to store `unique_ptr<AtHit>` instead of `AtHit`. Also move many instances of `std::vector<AtHit>` to `std::vector<const AtHit *>`. May change to `std::vector<unique_ptr<AtHit>>` in future to avoid a copy, but honestly I don't think it is necessary. The code is basically the same speed as before.

Refactor AtPSA so instead of implementing `void Analyze(AtEvent*, AtRawEvent*)` and being responsible for filling the entire event from a raw event, AtPSA implements `vector<unique_ptr<AtHit>> AnalyzePad(AtPad*)`. The base class handles setting the event metadata so it doesn't have to be re-implemented by each PSA.


anthoak13:
To the extent I can tell, things look the same.

image
image
image

Yassid:
Checked performance (seems good). I have not checked the fit performance but I assume nothing has been changed there in terms of variance calculation. It would be nice to have some documentation on the AtcontainerManip. It is a nice feature that should be used everywhere is needed.

PR #151, opened 2022-08-17
Cleanup AtTPC20MgDecay so it passes code linters

anthoak13:
@Yassid Don't worry about this PR, I'm working with Ruchi to cleanup the generator.


anthoak13:
I am making an executive decision to merge this in so our build output is not spammed by the issues fixed here. If it breaks something, we can revert and fix whatever the issue is.

PR #153, opened 2022-08-22
Specmat unpack alpha data

ACeulemans:
Update for specmat alpha source data analysis.
Updated the unpacking macro and added mapping.


AtMap/AtSpecMATMap.cxx
```diff
@@ -279,8 +283,29 @@ XYPoint AtSpecMATMap::CalcPadCenter(Int_t PadRef)
       return {-9999, -9999};
    }
 
-   auto x = AtPadCoord[PadRef][3][0];
-   auto y = AtPadCoord[PadRef][3][1];
+   double x, y;
```
anthoak13, line 286: x and y need to be given an initial value. Following the convention probably -9999

anthoak13 (changes requested):

AtMap/AtSpecMATMap.cxx
ACeulemans: Ok, I added the initialization for x & y.

PR #157, opened 2022-09-08
Implementation of a specific  e-field distortion correction

Yassid:


AtTools/AtEDistortionModel.cxx
```diff
@@ -0,0 +1,191 @@
...
+
+   auto [zcorr, radcorr, tracorr] = GetCorrectionFactors((Int_t)inputRad, (Int_t)inputZ);
+
+   Double_t outputRad = TMath::Sqrt(TMath::Power(inputRad + radcorr, 2) + TMath::Power(tracorr, 2));
+   Double_t outputAng = inputAng + TMath::ATan2(tracorr, inputRad + radcorr);
+
+   XYZPoint output;
+   output.SetX(outputRad * TMath::Cos(outputAng));
+   output.SetY(outputRad * TMath::Sin(outputAng));
+   output.SetZ(input.Z() - zcorr);
```
anthoak13, line 58: Can use GenVector again to do stuff like:
```
output.SetRho(outputRad)
output.SetPhi(outputAng)
```

AtTools/AtEDistortionModel.cxx
```diff
@@ -0,0 +1,191 @@
...
+
+   LOG(debug) << " Output point : " << output.X() << " " << output.Y() << " " << output.Z() << "\n";
+
+   return output;
+}
+
+XYZPoint AtEDistortionModel::ApplySpaceCharge(const XYZPoint &input)
+{
+   XYZPoint p;
```
anthoak13, line 67: I think you could shorten this to `return {};`

AtTools/AtEDistortionModel.cxx
```diff
@@ -0,0 +1,191 @@
...
+
+      std::cerr << " AtEDistortionModel::SetCorrectionMaps - Error! Missing LUT "
+                << "\n";
+      std::cerr << zlut << "\n";
+      std::cerr << radlut << "\n";
+      std::cerr << tralut << "\n";
+      return kFALSE;
+   }
+
+   return kTRUE;
```
anthoak13, line 99: unreachable code, return statements in both branches of if.

AtTools/AtEDistortionModel.cxx
```diff
@@ -0,0 +1,191 @@
...
+}
+
+std::tuple<Double_t, Double_t, Double_t>
+AtEDistortionModel::GetCorrectionFactors(const Int_t &radius, const Int_t &zpos)
```
anthoak13, line 103: Fine to take references, but is the same as just using an `Int_t radius` (addresses and ints are the same size).

This also doesn't matter, and is just a style thing, but you can just pass around EFieldCorr objects to avoid converting between those and tuples in the code.

AtTools/AtEDistortionModel.cxx
```diff
@@ -0,0 +1,191 @@
...
+      Double_t value;
+      std::istringstream buffer(line);
+      Int_t cnt = 0;
+
+      buffer >> radius;
+
+      while (buffer >> value) {
+
+         EFieldMapRef ref(radius, cnt);
+         map->emplace(ref, value);
```
anthoak13, line 140: Could probably replace with `map->emplace({radius, cnt}, value);` Little shorted and imo just as readable.

AtTools/AtEDistortionModel.cxx
```diff
@@ -0,0 +1,191 @@
...
+   std::cout << " Tra map size : " << tramap->size() << "\n";
+
+   // NB: This is not a necessary condition but it works for the present case
+   assert((zmap->size() == radmap->size()) && (radmap->size() == tramap->size()));
+
+   for (const auto &[key, zcorr] : *zmap) {
+      auto radcorr = radmap->at(key);
+      auto tracorr = tramap->at(key);
+      EFieldCorr corr(zcorr, radcorr, tracorr);
+      fDistortionMap->emplace(key, corr);
```
anthoak13, line 160: Similar to above comment, can use aggregate initialization (https://en.cppreference.com/w/cpp/language/aggregate_initialization).

AtTools/AtEDistortionModel.cxx
```diff
@@ -0,0 +1,191 @@
...
+      }
+   }
+}
+
+void AtEDistortionModel::SetGlobalMap(BufferMap *zmap, BufferMap *radmap, BufferMap *tramap)
+{
+
+   std::cout << " Z map size : " << zmap->size() << "\n";
+   std::cout << " Rad map size : " << radmap->size() << "\n";
+   std::cout << " Tra map size : " << tramap->size() << "\n";
```
anthoak13, line 151: couts should maybe be LOG(info)s?

AtTools/AtEDistortionModel.cxx
```diff
@@ -0,0 +1,191 @@
...
+   EFieldMapRef hashRef(200, 200);
+
+   std::cout << " Global map size : " << fDistortionMap->size() << "\n";
+   std::cout << " Hash value for (200,200)     : " << hashT(hashRef) << "\n";
+}
+
+void AtEDistortionModel::LoadParameters(AtDigiPar *par)
+{
+   if (par == nullptr)
+      return;
```
anthoak13, line 174: Doesn't actually do anything, can just be `{}`

AtTools/AtEDistortionModel.h
```diff
@@ -0,0 +1,90 @@
...
+class AtDigiPar;
+/**
+ * @brief Model for E-field distortion correction
+ *
+ *
+ *
+ *     The implemented correction is applied to every hit
+ *     Radial, transversal and longitudinal (z) corrections as a function of radial distance and z (map keys).
+ *     The Class EField mapping contains the correction values
+ */
```
anthoak13, line 21: This doxygen comment will refer just to the struct EFieldCorr, which I don't think is the intention? It seems to be describing the entire model so should be above the actual model class.

AtTools/AtEDistortionModel.h
```diff
@@ -0,0 +1,90 @@
...
+   inline void SetRadCorr(Double_t radcorr) { _radcorr = radcorr; }
+   inline void SetTraCorr(Double_t tracorr) { _tracorr = tracorr; }
+   inline Double_t GetZCorr() { return _zcorr; }
+   inline Double_t GetRadCorr() { return _radcorr; }
+   inline Double_t GetTraCorr() { return _tracorr; }
+
+   Double_t _zcorr;
+   Double_t _radcorr;
+   Double_t _tracorr;
+};
```
anthoak13, line 38: This has too much machinery around it that will keep users from using modern c++ features with structs, many of which are disabled when user-provided constructors (or default init values) are added and they are treated like classes instead of like structs. Should probably just be 
```
struct EFieldCorr {
Double_t zcorr;
Double_t radcorr;
Double_t tracorr;
};
```
(See AtPadReference for example)

AtTools/AtEDistortionModel.h
```diff
@@ -0,0 +1,90 @@
...
+
+namespace std {
+template <>
+struct hash<EFieldMapRef> {
+   inline size_t operator()(const EFieldMapRef &x) const { return x._rad + x._zpos * 10000; }
+};
+} // namespace std
+
+using eDistortionMap = std::unordered_map<EFieldMapRef, EFieldCorr>; // Global LUT
+using BufferMap = std::unordered_map<EFieldMapRef, Double_t>;        //! To parse individual LUT
```
anthoak13, line 61: I would put these in the AtDistortionModel class so they don't pollute the global namespace.

anthoak13 (changes requested):
There are a couple of style things to match the coding standards, and I think it would be useful to fully embrace using structs rather than half class/half structs for the internal data structures used in the distortion model.

AtGenerators/AtTPCIonDecay.cxx
```diff
@@ -256,9 +256,8 @@ Bool_t AtTPCIonDecay::ReadEvent(FairPrimaryGenerator *primGen)
       if (s > pow(M_tot, 2)) {
          // if(ExEject*1000.0>fSepEne){
          fIsDecay = kTRUE;
-         event1.SetDecay(fEnergyImpulsionLab_Total, fMult.at(Case), mass_1);
-         // Double_t weight1 = event1.Generate();
-
+         Bool_t decay = event1.SetDecay(fEnergyImpulsionLab_Total, fMult.at(Case), mass_1);
+         Double_t weight1 = event1.Generate();
```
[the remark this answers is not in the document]
Yassid: This is not part of the code I have modified

AtTools/AtEDistortionModel.cxx
Yassid: Wasnt able to compile it

anthoak13: For this to compiled the suggested changes to the struct had to be made to so it could use aggregate initialization https://en.cppreference.com/w/cpp/language/aggregate_initialization

Yassid: Thats what I actually did, but it didnt like it
141:36: error: no matching function for call to ‘std::unordered_map<EFieldMapRef, double>::emplace(<brace-enclosed initializer list>, Double_t&)’
  141 |     map->emplace({radius,cnt},value);

anthoak13: Someday I'll figure out why that is. I expected it to work like this `fAtMapPtr->AddAuxPad({10, 0, 0, 0}, "MCP_US");` but something must be different with the map emplace function signature. My guess not testing is emplace takes an r value, and it can't bind to whatever is actually being returned by the `{}` while in a the AddAuxPad it can bind to the const reference used there.

PR #158, opened 2022-09-25
 Implementation of a specific e-field distortion correction (refactored)

Yassid:


anthoak13:
I'm going to merge this and, then just clean up the clang-format errors in the commit I'm working on now.

anthoak13:
@Yassid You may have to do something like this https://github.com/ATTPC/ATTPCROOTv2/wiki/Collaborative-Git-Workflow#how-to-recover-after-upstream-branch-was-rebased or a hard reset on your develop branch to get it inline with upstream/develop again since it thinks you are 5 commits ahead and one behind even though you are even.

Yassid:
I am not sure why I have these problems with clang-format...
Also, not sure why I am 5 commits ahead and I am not sure if I can just discard them

anthoak13:
You should be safe to discard them, but I would probably quickly spin off a new branch called something like developBackup before trying to reset your develop to upstream develop. 

Your 5 commits ahead because of how swishing works. It's a rebase command, so it rewrites history. When we squash and merge a totally new commit is created that contains the contents of all the commits in the PR but there is no memory of where those commits came from. So from guys perspective there are now 6 unique commits, the original 5 and the merged 1. This why submitting PRs from feature branches is nice because after they're merged, the feature branches can be discarded basically.

PR #160, opened 2022-09-27
Electronic response

anthoak13:


anthoak13:
This doesn't change much structurally to the code. The main change to base classes is switching the response function used in AtPulseTask from a function pointer to a std::function so it is more general and can support both function pointers and callable objects. Allows for a channel-dependent response function to be used.

PR #161, opened 2022-09-28
Add FPN to HDF5 unpacker and relevant filters

anthoak13:
Also adds some macros for cleaning 22Mg data


Yassid:
Checked format and basic functionalities.

PR #167, opened 2022-10-10
Specmat simulations

ACeulemans:
Added a new simulation for the specmat detector for the Kr86(He4,He3)Kr87 reaction.


Yassid:
SpecMAT modifications. No major updates.

PR #169, opened 2022-10-31
Add support for when metadata is not saved in H5 file

anthoak13:
Will loop through each group in "get" and pull the min/max internal to MUTANT event number.


Yassid:
Tested

PR #174, opened 2022-11-21
Kine fitting

Yassid:
Simulation needs a major revision. Two body works as expected. Not sure about the rest. Added also kinematical  fitting, not tested yet.


Yassid:
Is profanity allowed in github?

Yassid:
Need to merge to progress on analysis. Couple of things need to be checked: At_d2He simulation has changed track ID assignment to be adapted to the new simulation scheme. Needs clang formatting (sowwy!)

PR #191, opened 2023-02-09
Various updates and additions to viewer

cnhunt:
Added functionality to AtViewerManager to use the isGood status of the raw event object to select events when using next and back [can be used with AtDataReductionTask to select specific events for viewing]

Added AtBoolMacro, which returns a bool from a user defined function using an AtTabInfo for event object access. [One application is use with AtDataReductionTask]

Minor fix to AtSidebarAddon to resize the frame when changing the string for an info box.

Minor fix to AtCutHEIST to catch an error state

### DIFF
--- a/AtEventDisplay/AtViewerManager.h
+++ b/AtEventDisplay/AtViewerManager.h
@@ -15,7 +15,7 @@
 #include <vector>
 
 class AtEventSidebar;
-class AtMap; // lines 29-29
+class AtMap;            // lines 29-29
 class AtTabBase;        // lines 31-31
 class FairTask;         // lines 19-19
 class TBuffer;          // lines 20-20


AtEventDisplay/AtViewerManager.h
```diff
@@ -69,13 +76,16 @@ class AtViewerManager final : DataHandling::AtObserver {
    DataHandling::AtTreeEntry &GetCurrentEntry() { return fEntry; }
    DataHandling::AtPadNum &GetPadNum() { return fPadNum; }
 
+   void SetCheckGood() { fCheckGood = true; }
+   void SetCheckBranch(TString name) { fCheckBranchName = name; }
+
```
anthoak13, line 81: Using info from above this would be
```cpp
void SetCheckBranch(AtBranch &branch) { fCheckGood = true; fCheckEvt = new FairRootEventInfo<AtRawEvent>(branch);}
```

macro/e12014/curtis/viewer/run_eve_dQdtA.C
```diff
@@ -0,0 +1,146 @@
...
+   cut->AddAllSpecies();
+   //cut->AddSpecies("Pb197");
+
+   auto boolMacEvt = std::make_unique<AtTabInfoFairRoot<AtRawEvent>>(eveMan->GetRawEventName());
+   AtBoolMacro *boolMac = new AtBoolMacro(TestTPC);
+   boolMac->GetTabInfo()->AddAugment(std::move(boolMacEvt));
+
+   AtDataReductionTask *reduceTask = new AtDataReductionTask();
+   reduceTask->SetInputBranch("AtRawEventSub");
+   reduceTask->SetReductionFunction([cut, boolMac]() { return (*cut)() && boolMac->ExecFunction(); });
```
anthoak13, line 102: This is the bit I referenced in removing AtBoolMacro since I think it adds complexity without adding ability to do things.

anthoak13 (changes requested):
Big note is I think AtBoolMacro can probably just be removed. Existing features in the language provide the same features with less lines of code/classes to maintain.

Also the tracking of the AtRawEvent to use for IsGoog should use the existing AtTabInfo classes (eventually I am going to replace all of the branch managment with that I think, there is a lot of boilerplate code dealing with the TClonesArray crap in FairRoot tasks as is)

AtEventDisplay/AtSidebar/AtSidebarAddon.cxx
```diff
@@ -72,8 +72,10 @@ void AtSidebarAddon::SetInfoString(std::string label, TString value)
 {
    if (fStrings.find(label) == fStrings.end())
       LOG(error) << label << " not defined!";
-   else
+   else {
       fStrings.find(label)->second->SetText(value);
+      dynamic_cast<TGFrame *>(const_cast<TGWindow *>(this->GetParent()))->Layout();
```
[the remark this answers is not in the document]
cnhunt: Because this->GetParent() returns a const TGWindow so it complains unless I cast that away first.

anthoak13 (approved):

PR #196, opened 2023-04-06
Add simple simulation task (bypass GEANT and use AtELossModels)

anthoak13:
Also allows for the sim and digi to happen in the same macro.

This is work moving towards developing a MC-fit like track reconstruction algorithm where energy loss models are more unknown then the lower energy/masses typically used by the AT-TPC.


Yassid:
Thanks, this was really needed. I want to add charge exchange cross sections to this scheme.

PR #221, opened 2023-10-17
Misc Additions

cnhunt:
Various additions and changes made while trying to calculate cross sections.

- Enhances the AtMacroTask to have functions run on the Init step
- Expanded simulation tools
     * More control to AtPulse and AtPulseTask
     * AtSimpleSimulation can now have custom conditions for the end of the track
     * Gaussian generator can now have an (X,Y) offset
- Added new AtSidebarIntMacro for creating custom sidebar fields with an integer i/o
- Added new AtCopyAuxTree to copy a tree from an auxiliary root file (such as from a separate DAQ) event by event based on if a AtRawEvent  object is good or not (to maintain parity during data reduction)

### DIFF
--- a/AtDigitization/AtPulse.h
+++ b/AtDigitization/AtPulse.h
@@ -63,6 +63,7 @@ public:
    void SetNoiseSigma(double val) { fNoiseSigma = val; }
    void SetSaveCharge(bool val) { fSaveCharge = val; }
    void SetDoConvolution(bool val) { fDoConvolution = val; }
+   void SetLowGain(double val) { fLowGainFactor = val; }
 
    AtRawEvent GenerateEvent(std::vector<SimPointPtr> &vec);
    AtRawEvent GenerateEvent(std::vector<AtSimulatedPoint *> &vec);
--- a/AtDigitization/AtPulse.h
+++ b/AtDigitization/AtPulse.h
@@ -41,6 +41,8 @@ protected:
    int fTBEntrance = 0;       //! Window location in timebuckets (from config)
    int fTBPadPlane = 0;       //! Pad plane location in TBs (calculated from DriftVelocity, TBEntrance, ZPadPlane
 
+   double fScaling = 1; // Scales the trace by this amount;
+
    ResponseFunc fResponse; //! Response function of the electronics
    bool fUseFastGain = true;
    double fNoiseSigma = 0; //! Sigma of random gaussian noise to apply to trace
@@ -64,6 +66,7 @@ public:
    void SetSaveCharge(bool val) { fSaveCharge = val; }
    void SetDoConvolution(bool val) { fDoConvolution = val; }
    void SetLowGain(double val) { fLowGainFactor = val; }
+   void SetScaling(double val) { fScaling = val; }
 
    AtRawEvent GenerateEvent(std::vector<SimPointPtr> &vec);
    AtRawEvent GenerateEvent(std::vector<AtSimulatedPoint *> &vec);
--- a/AtDigitization/AtPulse.h
+++ b/AtDigitization/AtPulse.h
@@ -41,8 +41,6 @@ protected:
    int fTBEntrance = 0;       //! Window location in timebuckets (from config)
    int fTBPadPlane = 0;       //! Pad plane location in TBs (calculated from DriftVelocity, TBEntrance, ZPadPlane
 
-   double fScaling = 1; // Scales the trace by this amount;
-
    ResponseFunc fResponse; //! Response function of the electronics
    bool fUseFastGain = true;
    double fNoiseSigma = 0; //! Sigma of random gaussian noise to apply to trace
@@ -66,7 +64,6 @@ public:
    void SetSaveCharge(bool val) { fSaveCharge = val; }
    void SetDoConvolution(bool val) { fDoConvolution = val; }
    void SetLowGain(double val) { fLowGainFactor = val; }
-   void SetScaling(double val) { fScaling = val; }
 
    AtRawEvent GenerateEvent(std::vector<SimPointPtr> &vec);
    AtRawEvent GenerateEvent(std::vector<AtSimulatedPoint *> &vec);
--- a/AtDigitization/AtPulseTask.h
+++ b/AtDigitization/AtPulseTask.h
@@ -12,6 +12,7 @@
 
 #include <Rtypes.h>
 #include <TClonesArray.h>
+#include <TString.h>
 
 #include <cstddef>
 #include <functional> // for function
--- a/AtDigitization/AtPulseTask.h
+++ b/AtDigitization/AtPulseTask.h
@@ -40,6 +40,8 @@ protected:
    Bool_t fIsPersistentAtTpcPoint{false}; //!< If true, save container
    Bool_t fSaveMCInfo{false};             //!<< Propagates MC information (adds AtTpcPoint branch to output
 
+   TString fOutputBranchName{"AtRawEvent"};
+
    TClonesArray *fSimulatedPointArray{nullptr}; //!< drifted electron array (input)
    TClonesArray *fMCPointArray{nullptr};        //!< MC Point Array (input)
    TClonesArray fRawEventArray;                 //!< Raw Event array (only one)
@@ -56,6 +58,7 @@ public:
    void SetPersistence(Bool_t val) { fIsPersistent = val; }
    void SetPersistenceAtTpcPoint(Bool_t val) { fIsPersistentAtTpcPoint = val; }
    void SetSaveMCInfo() { fSaveMCInfo = true; }
+   void SetOutputBranch(TString branchName) { fOutputBranchName = branchName; }
 
    virtual InitStatus Init() override;        //!< Initiliazation of task at the beginning of a run.
    virtual void Exec(Option_t *opt) override; //!< Executed for each event.
--- a/AtDigitization/AtSimpleSimulation.h
+++ b/AtDigitization/AtSimpleSimulation.h
@@ -12,6 +12,7 @@
 #include <TClonesArray.h>
 #include <TObject.h>
 
+#include <functional> // for function
 #include <map>
 #include <memory>
 #include <mutex>
@@ -64,7 +65,9 @@ public:
    void SetDistanceStep(double step) { fDistStep = step; } //<In mm
 
    void NewEvent();
-   void SimulateParticle(int Z, int A, const XYZPoint &iniPos, const PxPyPzEVector &iniMom);
+   std::pair<XYZPoint, PxPyPzEVector> SimulateParticle(
+      int Z, int A, const XYZPoint &iniPos, const PxPyPzEVector &iniMom,
+      std::function<bool(XYZPoint, PxPyPzEVector)> func = [](XYZPoint pos, PxPyPzEVector mom) { return true; });
 
    AtMCPoint &GetMcPoint(int i) { return dynamic_cast<AtMCPoint &>(*fMCPoints.At(i)); }
    int GetNumPoints() { return fMCPoints.GetEntries(); }
@@ -75,7 +78,9 @@ protected:
    bool IsInVolume(const std::string &volName, const XYZPoint &point);
    std::string GetVolumeName(const XYZPoint &point);
 
-   void SimulateParticle(ModelPtr model, const XYZPoint &iniPos, const PxPyPzEVector &iniMom);
+   std::pair<XYZPoint, PxPyPzEVector> SimulateParticle(
+      ModelPtr model, const XYZPoint &iniPos, const PxPyPzEVector &iniMom,
+      std::function<bool(XYZPoint, PxPyPzEVector)> func = [](XYZPoint pos, PxPyPzEVector mom) { return true; });
    void AddHit(double ELoss, const XYZPoint &pos, const PxPyPzEVector &mom, double length);
    TGeoVolume *GetVolume(const XYZPoint &pos);
 };
--- a/AtDigitization/AtSimpleSimulation.h
+++ b/AtDigitization/AtSimpleSimulation.h
@@ -16,7 +16,8 @@
 #include <map>
 #include <memory>
 #include <mutex>
-#include <string> // for string
+#include <string>  // for string
+#include <utility> // for pair
 namespace AtTools {
 class AtELossModel;
 }
@@ -68,9 +69,8 @@ public:
 
    /**
     * Simulates a particle over a given distance and returns the position and momentum of the particle at the stoping
-    * point. By default the particle will stop when it reaches the end of the TPC. A user defined function can test the
-    * position and momentum of the particle for each time step and stop it when a given condition is met (such as a
-    * depth in the TPC or an energy to stop at).
+    * point. Uses Z and A to provide a model to the protected version of SimulateParticle (see below for more
+    * information on the simulation).
     */
    std::pair<XYZPoint, PxPyPzEVector> SimulateParticle(
       int Z, int A, const XYZPoint &iniPos, const PxPyPzEVector &iniMom,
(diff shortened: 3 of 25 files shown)


anthoak13:
@cnhunt Can you add some doxygen comments to the simplesim stuff where you added inputs/outputs? I think I remember what you where doing but want to make sure I'm not misremembering before I do a deeper dive in.

AtDigitization/AtPulse.h
```diff
@@ -41,6 +41,8 @@ class AtPulse {
    int fTBEntrance = 0;       //! Window location in timebuckets (from config)
    int fTBPadPlane = 0;       //! Pad plane location in TBs (calculated from DriftVelocity, TBEntrance, ZPadPlane
 
+   double fScaling = 1; // Scales the trace by this amount;
+
```
anthoak13: What is the point of this? Doesn't it just scale the gain? Wouldn't it be simpler to adjust the gain parameter?

AtDigitization/AtSimpleSimulation.h
```diff
@@ -64,7 +65,16 @@ class AtSimpleSimulation {
...
+
+   /**
+    * Simulates a particle over a given distance and returns the position and momentum of the particle at the stoping
+    * point. By default the particle will stop when it reaches the end of the TPC. A user defined function can test the
+    * position and momentum of the particle for each time step and stop it when a given condition is met (such as a
+    * depth in the TPC or an energy to stop at).
+    */
+   std::pair<XYZPoint, PxPyPzEVector> SimulateParticle(
+      int Z, int A, const XYZPoint &iniPos, const PxPyPzEVector &iniMom,
+      std::function<bool(XYZPoint, PxPyPzEVector)> func = [](XYZPoint pos, PxPyPzEVector mom) { return true; });
```
anthoak13, line 77: This function should take const references, but fine as is with a copy (it's only 7 numbers).

anthoak13: The doxygen comment could also just reference the second `SimulateParticle` function, but say the difference (this requires we have added the model for this A/Z

AtDigitization/AtPulseTask.h
```diff
@@ -40,6 +40,8 @@ class AtPulseTask : public FairTask {
    Bool_t fIsPersistentAtTpcPoint{false}; //!< If true, save container
    Bool_t fSaveMCInfo{false};             //!<< Propagates MC information (adds AtTpcPoint branch to output
 
+   TString fOutputBranchName{"AtRawEvent"};
```
anthoak13, line 44: Need to add header for TString.

AtDigitization/AtSimpleSimulation.h
```diff
@@ -12,6 +12,7 @@
 #include <TClonesArray.h>
 #include <TObject.h>
 
+#include <functional> // for function
 #include <map>
 #include <memory>
 #include <mutex>
```
anthoak13, line 18: Need to add `<utility>` for `std::pair`

AtEventDisplay/AtSidebar/AtSidebarIntMacro.h
```diff
@@ -0,0 +1,54 @@
...
+#include <functional>
+#include <string>
+#include <utility> // for move
+
+class TBuffer;
+class TClass;
+class TGWindow;
+class TMemberInspector;
+
+class AtSidebarIntMacro : public AtSidebarAddon {
```
anthoak13, line 19: Needs doxygen. What does this class do? This feels like one of the sidebar helper things? I think those are all in the same file. Is there a reason for this to not also be there?

AtEventDisplay/AtSidebar/AtSidebarIntMacro.h
```diff
@@ -0,0 +1,54 @@
...
+private:
+   using MacroFunction = std::function<void(int)>;
+   MacroFunction fFunction;
+
+   std::string fLabel;
+
+   int fMin{0};
+   int fMax{10};
+   int fStart{0};
```
anthoak13: These are determined at creation time? Is there a use case where they would be specified separately or can they be added to the constructor and not stored as part of the state of the class?

AtReconstruction/AtMacroTask.h
```diff
@@ -9,20 +9,29 @@
 class TClass;
```
anthoak13, line 11:
```
/user/anthonya/attpcroot/AtReconstruction/AtMacroTask.h should add these lines:
#include <Rtypes.h>    // for THashConsistencyHolder, ClassDef, Option_t
#include <algorithm>   // for max
class TBuffer;
class TMemberInspector;
```

AtReconstruction/AtCopyAuxTreeTask.h
```diff
@@ -0,0 +1,77 @@
...
+ * corresponding events to a new output file, ensuring parity between the root files
+ * from the two DAQs.
+ *
+ */
+
+#include <TChain.h>
+#include <TString.h>
+#include <TTree.h>
+// FairRoot classes
+#include <FairTask.h>
```
anthoak13, line 18: /user/anthonya/attpcroot/AtReconstruction/AtCopyAuxTreeTask.h should remove these lines:
- #include <TTree.h>  // lines 16-16
- #include <vector>  // lines 23-23

AtReconstruction/AtCopyAuxTreeTask.cxx
```diff
@@ -0,0 +1,89 @@
+#include "AtCopyAuxTreeTask.h"
+
+#include "AtRawEvent.h"
+
+#include <FairLogger.h>
```
anthoak13, line 5: /user/anthonya/attpcroot/AtReconstruction/AtCopyAuxTreeTask.cxx should add these lines:
#include <TFile.h>            // for TFile
#include <TTree.h>            // for TTree

AtReconstruction/AtCopyAuxTreeTask.cxx
```diff
@@ -0,0 +1,89 @@
...
+#include <FairRootFileSink.h>
+#include <FairRootManager.h>
+#include <FairRunAna.h>
```
anthoak13: /user/anthonya/attpcroot/AtReconstruction/AtCopyAuxTreeTask.cxx should remove these lines:
- #include <FairRootFileSink.h>  // lines 6-6
- #include <FairRunAna.h>  // lines 8-8
- #include <FairSink.h>  // lines 9-9

AtEventDisplay/AtSidebar/AtSidebarIntMacro.cxx
```diff
@@ -0,0 +1,18 @@
+#include "AtSidebarIntMacro.h"
+
+#include <FairLogger.h> // for endl, basic_ostream, cout, ostream
```
anthoak13: /user/anthonya/attpcroot/AtEventDisplay/AtSidebar/AtSidebarIntMacro.cxx should remove these lines:
- #include <FairLogger.h>  // lines 3-3

AtReconstruction/AtCopyAuxTreeTask.cxx
```diff
@@ -0,0 +1,89 @@
...
+   if (ioMan == nullptr) {
+      LOG(fatal) << "Cannot find RootManager!";
+      return kFATAL;
+   }
+
+   fRawEventArray = dynamic_cast<TClonesArray *>(ioMan->GetObject(fRawEventBranchName));
+   if (fRawEventArray == nullptr) {
+      LOG(fatal) << "Cannot find AtRawEvent array in branch " << fRawEventBranchName << "!";
+      return kFATAL;
+   }
```
anthoak13, line 53: Why is there here? Is there a reason to require a raw event is in the tree?

anthoak13: This should be an AtBaseEvent, not and AtRawEvent. No reason to require the user to have a raw event in the ttree when all we are doing is looking at if an event is good or not.

AtTools/AtLineChargeModel.cxx
```diff
@@ -49,6 +49,10 @@ XYZPoint AtLineChargeModel::ApplySpaceCharge(const XYZPoint &reverseInputPositio
    auto delZ = input.Z();         // in mm
    double dist2 = getDist2(delZ); // in mm;
    double newRho = sqrt(input.Rho() * input.Rho() - dist2);
+   if (!(newRho > 0)) {
```
anthoak13: Easier to understand if code is:
```
if(newRho < 0) newRho = 0;
```

It took me longer than it should to decode this lol

anthoak13 (changes requested):

AtDigitization/AtPulse.h
cnhunt: The gain is taken from the parameter file. An adjustable scaling factor is required because our measured response function isn't perfectly normalized. It could be done with gain, but would require either exposing the gain to be a user defined value or making a different parameter file depending on the used response function, neither of which seem appropriate.

AtDigitization/AtSimpleSimulation.h
cnhunt: I'm not sure I agree with the doxygen comment since one function is public and the other is protected.

AtEventDisplay/AtSidebar/AtSidebarIntMacro.h
cnhunt: AtSidebarIntMacro is like AtSidebarInfoMacro. Both of these are generic sidebar items that can be customized. In this case it is an int box that can be used to do arbitrary things. It means we don't need a special class for each thing we might want adjustable parameters for in the viewer, like we have for some of the PSA methods.

AtDigitization/AtPulse.h
anthoak13: If it's to compensate for differences in the response function, that feels like it should be folded into that rather than a misc other parameter in AtPulseTask. I'd prefer a response-function dependent gain parameter over another random scaling factor.

AtDigitization/AtSimpleSimulation.h
anthoak13: My concern is someone changes some logic somewhere and doesn't update both dioxygen comments. I'd be fine with the second (protected) referencing the public too. I just don't like the duplication of information for long-term maintenance.

AtEventDisplay/AtSidebar/AtSidebarIntMacro.h
anthoak13: This might be beyond what you want to do, but what I think this should be replaced with eventually:

Using signals/slots https://root.cern/manual/signal_slot/ it should be possible to add this capability into AtSIdebarAddon where the callback includes the label of the entry that we can use to call the correct MacroFunction. (AtSidebarAddaon would require another `map<String, macropointer>`, but it would let us add multiple without spawning a new class for each one. 

If you don't want to do this now, we can still merge this in but it will (eventually) be replaced with this behavior if we can get it to work (it's more general and easier to use I think).

anthoak13: Right now. we need to instantiate a new class which adds a new sidebar object for each intbox. And it would be nice if we could just do the `AddIntBox(label, ... , std::func<void(int)>)`

cnhunt: I like this idea. But it's not happening before the holidays and this pull request has been sitting for long enough, I think.

I might take a look at adding that functionality once I'm back to work in the new year, but as it's own thing.

cnhunt: For now I added documentation to the class (and InfoMacro).

AtDigitization/AtPulse.h
cnhunt: Where might that go?

anthoak13: What response function is the problem?

anthoak13: Could add it to the base class https://github.com/ATTPC/ATTPCROOTv2/blob/develop/AtTools/ElectronicResponse/AtElectronicResponse.h.

If could be enforced by making the virtual functions protected and have non-virtual public classes that return the result of the protected * scalingFactor.

anthoak13: Actually the class is already setup that way so this:
```
class AtElectronicResponse {
protected:
   virtual double GetResponse(double time) const = 0;
   virtual double GetResponse(int padNum, double time) const = 0;

public:
   virtual ~AtElectronicResponse() = default;
   /**
    * Returns the response of the electronics to a delta-function of charge at t=0 after time (us).
    */
   double operator()(double time) const { return GetResponse(time); }
   /**
    * Returns the response of the electronics to a delta-function of charge at t=0 after time (us).
    */
   double operator()(int padNum, double time) const { return GetResponse(padNum, time); }
};
```

Becomes

```
class AtElectronicResponse {
private:
double  fScale{1};
protected:
   virtual double GetResponse(double time) const = 0;
   virtual double GetResponse(int padNum, double time) const = 0;

public:
   virtual ~AtElectronicResponse() = default;
   /**
    * Returns the response of the electronics to a delta-function of charge at t=0 after time (us).
    */
   double operator()(double time) const { return GetResponse(time) * fScale; }
   /**
    * Returns the response of the electronics to a delta-function of charge at t=0 after time (us).
    */
   double operator()(int padNum, double time) const { return GetResponse(padNum, time) * fScale; }
void SetScale(double scale) { fScale = scale;}

};

Comment (anthoak13) [reply AtDigitization/AtPulse.h]: Then issue sorted without adding more random stuff to AtPulseTask.

File: AtReconstruction/AtCopyAuxTreeTask.cxx (hunk 12 above)
Comment (cnhunt) [reply AtReconstruction/AtCopyAuxTreeTask.cxx:53]: This has been changed to an AtBaseEvent, now (didn't realize AtBaseEvent had the IsGood checking in it. This is definitely a good change)

File: AtReconstruction/AtMacroTask.h (hunk 7 above)
Comment (cnhunt) [reply AtReconstruction/AtMacroTask.h:11]: Why?

File: AtDigitization/AtSimpleSimulation.h (hunk 2 above)
Comment (cnhunt) [reply AtDigitization/AtSimpleSimulation.h:77]: I see what you mean. I think I've changed it to remove that potential future error.

File: AtReconstruction/AtMacroTask.h (hunk 7 above)
Comment (cnhunt) [reply AtReconstruction/AtMacroTask.h:11]: if Rtypes is required for ClassDef, I can see that. No idea why the rest would be relevant.

File: AtDigitization/AtPulse.h (hunk 1 above)
Comment (cnhunt) [reply AtDigitization/AtPulse.h]: Honestly, I don't think a lot of the response function stuff existed when I was working on this. I was working out of a root file.

EDIT: I see it did, but I never used it

File: AtEventDisplay/AtSidebar/AtSidebarFrames.cxx [hunk 14]
```diff
@@ -43,8 +43,11 @@ void AtSidebarRunInfo::FillFrame()
...
-
-   fRunFile = new TGLabel(this, (Infile + file->GetName()).Data());
+   if (file != nullptr) {
+      fRunFile = new TGLabel(this, (Infile + file->GetName()).Data());
+   } else {
+      fRunFile = new TGLabel(this, "null");
+   }
 
    // If the file path is too long to fit in the sidebar, just grab the file name
    if (fRunFile->GetWidth() > this->GetWidth()) {
```
anthoak13: This if statement should also live in the `if(file != nullptr)` statement since `file` can be null here

AtReconstruction/AtMacroTask.h
anthoak13: TBuffer and TMemberInspector are both used in the classdef and need to be forward declared to make IWYU happy so we are not spammed by every single TObject class. (Blame ROOT)

anthoak13 (approved):

PR #224, opened 2024-01-18
Analysis

Yassid:
Fitting is now a task. Brho calculation has been updated to use the result from the circle fit. Many analysis macros have been updated.


AtData/AtFittedTrack.h
```diff
@@ -0,0 +1,144 @@
...
+#include <iostream>
+#include <memory>
+#include <utility>
+#include <vector>
+
+class TBuffer;
+class TClass;
+class TMemberInspector;
+
+class AtFittedTrack : public TObject {
```
anthoak13, line 22: In the future I will look at changing this class to rely less on the tuples and generalize it to apply to both common fitting methods (MC and Genfit). Over the last FRIB experiment, I started this work but like we discussed it is not worth waiting for.

anthoak13 (approved):

anthoak13:
What version of clang-format are you using? I'm fine with these errors (they're just parameter pack things)

Yassid:
clang-format version 10.0.0-4ubuntu1

PR #250, opened 2025-06-10
Implemented the 3D Bragg curve fitter using CATima

RealAurio:
Been working on this feature. Finally managed to get something that seems to work. There is still a lot to fix, and probably a refactoring will be done once I have everything more clear. However, I wanted to create this pull request so that we can check how does this feature performs with the simulations that are being prepared by others in their respective forks.

CATima repo: https://github.com/hrosiak/catima

Notes:
1. In order to get this to work, I needed to compile CATima using the DGSL integration option (cmake -DCMAKE_INSTALL_PREFIX=~/fair_install/catimaInstall -DGSL_INTEGRATION=ON ..).
2. I also needed to comment 2 lines in the file "src/storage.cpp" file from CATima (in deleter function InterpolatorGSL::~InterpolatorGSL(), just comment both lines). Maybe it was just me, but it was giving me a segmentation fault.
3. Once you have installed CATima, export the environment variable CATIMA and set it to the path where you have installed CATima (it should have 3 folders, namely: include, lib and share). In my case: export CATIMA=~/fair_install/catimaInstall
4. Now simply compile ATTPCROOTv2. It should find CATima and therefore compile the At3DBraggFitter class.
5. You can test that it works by running the simulation in "macro/Simulation/ATTPC/12Be_dp".
6. I have removed GenFit from the AtFitter base class, since I checked it and it really doesn't use anything from GenFit. I initially wanted At3DBraggFitter to derive from AtFitter, so that is why I made that change. I hope it doesn't break anything.
7. Also, there are a couple of changes regarding the S800, since I was trying to test this with the data from the e23031 experiment. Sorry for mixing things.


AtData/At3DBraggFitResult.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <Rtypes.h>
+#include <TObject.h>
+
+#include <iostream>
+#include <memory>
+#include <utility>
+#include <vector>
+
+class At3DBraggFitResult : public TObject {
```
anthoak13, line 12: This class is similar in scope to the AtMCResult class that exists. I think we should talk about how to merge these two - maybe this derives from AtMCResult and that gets refactored to AtFitResult?

This and the MC fitting are basically doing the same thing - only one more than once per species.

AtData/AtTrack.h
```diff
@@ -44,6 +45,11 @@ class AtTrack : public TObject {
    std::pair<Double_t, Double_t> fGeoCenter; //< Center of the spiral track
    std::vector<AtHitCluster> fHitClusterArray; //< Clusterized hits container. Can also be stored in fHitArray
 
+   // Stuff related to the 3DBragg curve. May be moved to a child class in the future.
+   // Vector of pair of values for the 3DBragg curve of the track (archLength[mm], eLoss[a.u.]).
+   std::vector<std::pair<Double_t, Double_t>> f3DBraggCurveValues;
+   std::unique_ptr<At3DBraggFitResult> fFitResult{};
```
anthoak13, line 51: I don't like this being in the AtTrack class given it's specific to one of the three fitting methods we posses.

anthoak13: As a first thought, I would have the fit result as a separate branch on the tree. You can just save the FitResult class and store the track ID there for linking.

AtData/AtTrack.h
```diff
@@ -44,6 +45,11 @@ class AtTrack : public TObject {
    std::pair<Double_t, Double_t> fGeoCenter; //< Center of the spiral track
    std::vector<AtHitCluster> fHitClusterArray; //< Clusterized hits container. Can also be stored in fHitArray
 
+   // Stuff related to the 3DBragg curve. May be moved to a child class in the future.
+   // Vector of pair of values for the 3DBragg curve of the track (archLength[mm], eLoss[a.u.]).
+   std::vector<std::pair<Double_t, Double_t>> f3DBraggCurveValues;
```
anthoak13, line 50: Is this just a copy of the fit result?

AtEventDisplay/AtTabs/AtTab3DBraggCurve.h
```diff
@@ -0,0 +1,110 @@
...
+   TCanvas *fCvsDeDx{nullptr};
+   TMultiGraph *fDeDxGraphs{nullptr};
+
+   TCanvas *fCvsQvS{nullptr};
+   TH1F *fHistQvS{nullptr};
+   TGraph *fFittedBraggCurveGraph{nullptr};
+   Double_t fDZ{1};
+   Double_t fDS{1};
+
+   std::unique_ptr<AtFITTER::At3DBraggFitter> fFitter{nullptr};
```
anthoak13, line 61: Why does the tab have ownership over the fitter object?

AtEventDisplay/AtTabs/AtTab3DBraggCurve.h
```diff
@@ -0,0 +1,110 @@
...
+   void Exec() override{};
+   void Update(DataHandling::AtSubject *sub) override;
+
+   void SetThreshold(Int_t val) { fThreshold = val; }
+   void SetHitAttributes(TAttMarker attr) { fHitAttr = std::move(attr); }
+   void SetMultiHit(Int_t hitMax) { fMaxHitMulti = hitMax; }
+
+   void SetDZ(Double_t DZ) { fDZ = DZ; }
+
+   void Set3DBraggCurveFitter(std::unique_ptr<AtFITTER::At3DBraggFitter> fitter) { fFitter = std::move(fitter); }
```
anthoak13, line 81: If the fitter is required for the tab to function, it would make more sense to pass it in the constructor

AtEventDisplay/AtTabs/AtTab3DBraggCurve.h
```diff
@@ -0,0 +1,110 @@
...
+class TCanvas;
+class TH1F;
+namespace DataHandling {
+class AtSubject;
+}
+
+/*
+   Tab for 3D visualizing of events together with the 3DBragg curves of straight tracks.
+*/
+class AtTab3DBraggCurve : public AtTabBase, public DataHandling::AtObserver {
```
anthoak13, line 33: This tab seems to re-implement a large amount of the main tab? Or at least is re-implementing a lot of the same methods. Is there a reason for this? Would it make more sense to just inherit from AtTabMain?

anthoak13: Or are a lot of the functions not used?

AtEventDisplay/AtTabs/AtTab3DBraggCurve.cxx
```diff
@@ -0,0 +1,430 @@
...
+
+constexpr auto cRED = "\033[1;31m";
+constexpr auto cYELLOW = "\033[1;33m";
+constexpr auto cNORMAL = "\033[0m";
+constexpr auto cGREEN = "\033[1;32m";
+constexpr auto cBLUE = "\033[1;34m";
+
+ClassImp(AtTab3DBraggCurve);
+
+AtTab3DBraggCurve::AtTab3DBraggCurve() : AtTabBase("3DBraggCurve")
```
anthoak13, line 62: Given the amount of duplicate code, this needs to be based off of AtTabMain, rather than copy/paste. It will become difficult to maintain if we have multiple copies of the code (and duplicate events in the manager at once)

AtEventDisplay/AtTabs/AtTab3DBraggCurve.cxx
```diff
@@ -0,0 +1,430 @@
...
+   fCvsQvS->GetListOfPrimitives()->Remove(fHistQvS);
+   fCvsQvS->GetListOfPrimitives()->Remove(fFittedBraggCurveGraph);
+   if (tracks.size()) {
+      fTrackIdx = 0;
+      Draw3DBraggCurve(tracks[fTrackIdx]);
+   } else {
+      fTrackIdx = -1;
+      DrawDeDxGraphs();
+      DrawHistQvS();
+   }
```
anthoak13, line 196: It looks like this is the only new code? Method should call parent, then run this bit

AtEventDisplay/AtTabs/AtTab3DBraggCurve.cxx
```diff
@@ -0,0 +1,430 @@
...
+
+      auto position = hit.GetPosition();
+      hitSet.SetNextPoint(position.X() / 10., position.Y() / 10., position.Z() / 10.); // Convert into cm
+      hitSet.SetPointId(new TNamed(Form("Hit %d", i), ""));
+   }
+
+   gEve->ElementChanged(&hitSet);
+}
+
+void AtTab3DBraggCurve::MakeTab(TEveWindowSlot *slot)
```
anthoak13, line 267: This is the bit that might just fully override AtMainTab.

But I don't think this class does much besides display some historgrams? At least from what I have seen so far. It might make sense for this *not* to be a whole replacement for the main tab and just a normal tab that draws the histograms

AtEventDisplay/AtTabs/AtTab3DBraggCurve.cxx
```diff
@@ -0,0 +1,430 @@
...
+         fFittedBraggCurveGraph->AddPoint(pair.first, amplitudeFactor * pair.second);
+      fFittedBraggCurveGraph->Draw("same");
+   }
+
+   DrawDeDxGraphs();
+   if (fitResult != nullptr && fFitter != nullptr) {
+
+      for (Int_t i = 0; i < fitResult->GetNumberOfEntries(); i++) {
+         fFitter->SetProjectileIndex(i);
+         auto dEdx = fFitter->GetDeDx(fitResult->GetKineticEnergy(i));
```
anthoak13, line 416: Why are you making a full fitter object for this? Is this just getting the dedx?

AtReconstruction/At3DBraggCurveTask.h
```diff
@@ -0,0 +1,57 @@
...
+#include <Rtypes.h> // for Int_t, Bool_t, Double_t, THashConsistencyHolder
+#include <TClonesArray.h>
+#include <TString.h> // for TString
+
+class AtRawEvent;
+class AtPatternEvent;
+class TBuffer;
+class TClass;
+class TMemberInspector;
+
```
anthoak13, line 18: This needs comments. What is the output? Is it just a copy of the track with the extra members filled? Why not just output the fit results themselves so the data isn't coppied?

AtReconstruction/At3DBraggCurveTask.h
```diff
@@ -0,0 +1,57 @@
...
+   TClonesArray *fPatternEventArray{nullptr};
+   TClonesArray fOutputPatternEventArray;
+
+   AtRawEvent *fRawEvent{};
+   AtPatternEvent *fPatternEvent{};
+
+   Bool_t kIsPersistence;
+   Double_t fChargeThres{-1};
+
+   At3DBraggCurveHandler *f3DBraggHandler{nullptr};
```
anthoak13, line 35: Who owns this. Raw pointers should document ownership in comments

AtReconstruction/At3DBraggCurveTask.cxx
```diff
@@ -0,0 +1,93 @@
...
+
+   fPatternEvent = dynamic_cast<AtPatternEvent *>(fPatternEventArray->At(0));
+
+   fOutputPatternEventArray.Delete();
+   auto patternEvent3DBraggCurve = dynamic_cast<AtPatternEvent *>(fOutputPatternEventArray.ConstructedAt(0));
+
+   // We first add the noise.
+   auto &noiseHits = fPatternEvent->GetNoiseHits();
+   for (auto &&noiseHit : noiseHits)
+      patternEvent3DBraggCurve->AddNoise(std::move(*noiseHit));
```
anthoak13, line 78: Why is the noise added separately? That's already contained in the track object passed below, no?

This seems like logic that belongs to the fitter class, not the task class. The fitter class should work more or less as a stand-alone

AtReconstruction/At3DBraggCurveTask.cxx
```diff
@@ -0,0 +1,93 @@
...
+
+   std::vector<AtTrack> &tracks = fPatternEvent->GetTrackCand();
+   for (int i = 0; i < tracks.size(); ++i) {
+      AtTrack newTrack = AtTrack(tracks[i]);
+      if (fRawEventArray == nullptr)
+         f3DBraggHandler->SetTrack(newTrack);
+      else
+         f3DBraggHandler->SetTrack(newTrack, fRawEvent);
```
anthoak13, line 86: Also redundant logic? Presumably SetTrack is checking for null?

AtReconstruction/AtPatternRecognition/At3DBraggCurveHandler.h
```diff
@@ -0,0 +1,36 @@
...
+#include <Math/Point3D.h>
+#include <TH1F.h>
+
+#include <utility>
+#include <vector>
+
+class AtTrack;
+class AtRawEvent;
+
+class At3DBraggCurveHandler {
```
anthoak13, line 17: I don't love this class. It's worth talking about what problem it is trying to solve and how best to handle that.

AtReconstruction/AtFitter/At3DBraggFitter.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include <catima/catima.h>
+
+namespace AtFITTER {
+
+static TH1F *gHistExpBraggCurve{nullptr};
+static catima::Material *gMaterial{nullptr};
+static catima::Projectile *gProjectile{nullptr};
+static double gMassUma{1};
+static double gDS{1};
```
anthoak13, line 20: Why are these global? Should they be `static thread_local`? Or are all these classes thread safe?

AtReconstruction/AtFitter/At3DBraggFitter.h
```diff
@@ -0,0 +1,74 @@
...
+
+class At3DBraggFitter : public AtFitter {
+protected:
+   int fParticleIdx{0};
+   std::vector<std::pair<Int_t, Int_t>> fAZPairs{};
+   std::vector<double> fParticleMassesUma{};
+
+   Double_t fMaterialDensity{0};
+   std::vector<std::tuple<Int_t, Int_t, Int_t>> fMaterialComponents{};
```
anthoak13, line 29: Most of these needs comments to describe

AtReconstruction/AtFitter/At3DBraggFitter.h
```diff
@@ -0,0 +1,74 @@
...
+
+   double fDZ{1};
+   double fMinKE{0.5};
+   double fMaxKE{15};
+   double fKEGuess{2};
```
anthoak13, line 34: Same here (Doxygen comments)

AtReconstruction/AtFitter/At3DBraggFitter.h
```diff
@@ -0,0 +1,74 @@
...
+
+public:
+   At3DBraggFitter() = default;
+   ~At3DBraggFitter() = default;
+
+   void Init() override;
+
+   std::vector<std::unique_ptr<AtFittedTrack>> ProcessTracks(std::vector<AtTrack> &tracks) override;
```
anthoak13, line 42: This produces AtFittedTracks, but it seemed like most of the test of your code just worked with AtTracks?

AtReconstruction/AtFitter/At3DBraggFitter.h
anthoak13: Reading through the header, I think I've become convinced these should live within your class.

AtReconstruction/AtFitter/At3DBraggFitter.h
```diff
@@ -0,0 +1,74 @@
...
+
+   std::unique_ptr<At3DBraggFitResult> ProcessTrack(AtTrack &track);
+
+   void AddAZPair(std::pair<Int_t, Int_t> pair) { fAZPairs.push_back(pair); }
+   void AddMassUma(double value) { fParticleMassesUma.push_back(value); }
+
+   void SetMaterialDensity(Double_t value) { fMaterialDensity = value; }
+   void AddMaterialComponent(std::tuple<Int_t, Int_t, Int_t> component) { fMaterialComponents.push_back(component); }
+   void SetProjectileIndex(Int_t i);
```
anthoak13, line 51: Need documenting. Especially projectile index. It's not obvious what that function does and what secondary effects it may have.

AtReconstruction/AtFitter/At3DBraggFitter.h
```diff
@@ -0,0 +1,74 @@
...
+
+   void SetDZ(double value) { fDZ = value; }
+   void SetMinKE(double value) { fMinKE = value; }
+   void SetMaxKE(double value) { fMaxKE = value; }
+   void SetKEGuess(double value) { fKEGuess = value; }
+
+   static std::vector<std::pair<double, double>> GetDeDx(Double_t kineticEnergy);
+   static std::vector<std::pair<double, double>> GetELossModel(Double_t kineticEnergy);
```
anthoak13, line 59: These are out of place. They belong to the energy loss model, not the fitting class.

CATIMA should be connected to the existing energy loss model code ([AtELossModel](https://github.com/ATTPC/ATTPCROOTv2/blob/develop/AtTools/AtELossModel.h)) so its usable by all parts of the code base, not just this one class. This class should probably work of the generalized ELossModel too so it's decoupled from CATIMA in case other energy loss models are ever used.

anthoak13: This would also remove the hard dependency on CATIMA, and that CATIMA specific stuff would all be wrapped up into a few classes that are easy to turn on/off in AtTools

anthoak13: At the moment, catima has spread enough through the build system that it fails to build if it isn't installed (the build check failed for the PR)

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+
+   AtFITTER::gMaterial = new catima::Material();
+   AtFITTER::gMaterial->density(fMaterialDensity);
+
+   for (auto materialComponent : fMaterialComponents)
+      AtFITTER::gMaterial->add_element(std::get<0>(materialComponent), std::get<1>(materialComponent),
+                                       std::get<2>(materialComponent));
+}
+
+std::unique_ptr<At3DBraggFitResult> AtFITTER::At3DBraggFitter::ProcessTrack(AtTrack &track)
```
anthoak13, line 26: From reading, none of those global variables should exist!

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+{
+
+   gSystem->Load("libMinuit");
```
anthoak13, line 29: Minuit is fine for now - in future probably easier to use the [RooFit](https://root.cern/manual/roofit/) or [ROOT::Fit](https://root.cern/manual/fitting/#using-rootfit-classes) classes. Those are thread safe and minuit is very much NOT thread safe.  Those are thread safe. [Here is an example](https://github.com/ATTPC/ATTPCROOTv2/blob/0079ea48717d0c74eb109124fd6c6cebd34bf007/AtReconstruction/AtPulseAnalyzer/AtPSADeconvFit.cxx#L99) from our code of thread-safe fitting using minuit2 in the ROOT::Fit classes

anthoak13: [Here is an example](https://github.com/ATTPC/ATTPCROOTv2/blob/0079ea48717d0c74eb109124fd6c6cebd34bf007/AtReconstruction/AtFitter/AtMCFission.cxx#L154) using the ROOT::Fit classes with a custom objective function

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+
+   // Initialize an At3DBraggResult object to store the fit results.
+   std::unique_ptr<At3DBraggFitResult> fitResult = std::make_unique<At3DBraggFitResult>();
+
+   while (fParticleIdx < fAZPairs.size()) {
+
+      auto AZPair = fAZPairs[fParticleIdx];
+      AtFITTER::gMassUma = fParticleMassesUma[fParticleIdx];
+
+      catima::Projectile *p = new catima::Projectile(AZPair.first, AZPair.second);
```
anthoak13, line 63: Memory leak  - I didn't see a delete. Should not use raw pointer

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+      AtFITTER::gHistExpBraggCurve->Fill(pairBragg.first, pairBragg.second);
+
+   for (Int_t i = 1; i <= AtFITTER::gHistExpBraggCurve->GetNbinsX(); i++)
+      // AtFITTER::gHistExpBraggCurve->SetBinError(i, TMath::Sqrt(AtFITTER::gHistExpBraggCurve->GetBinContent(i)));
+      AtFITTER::gHistExpBraggCurve->SetBinError(i, 0.1 * AtFITTER::gHistExpBraggCurve->GetBinContent(i));
+
+   // Initialize an At3DBraggResult object to store the fit results.
+   std::unique_ptr<At3DBraggFitResult> fitResult = std::make_unique<At3DBraggFitResult>();
+
+   while (fParticleIdx < fAZPairs.size()) {
```
anthoak13, line 58: Everything in this while loop should be in its own (protected?) function. It's difficult to read as is.

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+            std::cout << "modelPair.first = " << modelPair.first << std::endl;*/
+            if (std::abs(modelPair.first - range) < 0.0001) {
+               // if (modelPair.first == range) {
+               // std::cout << "Hewo! :3" << std::endl;
+               double model = amplitudeFactor * modelPair.second;
+               if (err > 0)
+                  fval += pow((charge - model) / err, 2);
+               iteModel = j;
+               break;
+            }
```
anthoak13, line 152: I'm not sure I understand this

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+
+   fParticleIdx = 0;
+   return std::move(fitResult);
+}
+
+std::vector<std::unique_ptr<AtFittedTrack>> AtFITTER::At3DBraggFitter::ProcessTracks(std::vector<AtTrack> &tracks)
+{
+   std::vector<std::unique_ptr<AtFittedTrack>> letMeCompilePls;
+   return letMeCompilePls;
+}
```
anthoak13, line 186: This class should not inherit from AtFitter if it isn't going to follow AtFitter's contract!

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+
+void AtFITTER::At3DBraggFitter::SetProjectileIndex(Int_t i)
+{
+
+   auto AZPair = fAZPairs[i];
+   catima::Projectile *p = new catima::Projectile(AZPair.first, AZPair.second);
```
anthoak13, line 192: memory leak! When is this function used? It feels internal and I think I read this code in the large method above already

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+Double_t AtFITTER::At3DBraggFitter::EstimateEnergyFromRange(Double_t range)
+{
+
+   Double_t kineticEnergy{0.1};
+
+   AtFITTER::gProjectile->T = kineticEnergy / AtFITTER::gMassUma;
+   Double_t catimaRange =
+      catima::range(*AtFITTER::gProjectile, *AtFITTER::gMaterial) / AtFITTER::gMaterial->density() * 10;
+
+   while (catimaRange < range) {
```
anthoak13, line 206: Works but slow!

anthoak13: linear search

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+         }
+      } // For loop model values.
+   }    // For loop experimental values.
+
+   if (entries)
+      return estimatedAmplitudeFactor / entries;
+   return 3e3;
+}
+
+std::vector<std::pair<Double_t, Double_t>> AtFITTER::At3DBraggFitter::GetDeDx(Double_t kineticEnergy)
```
anthoak13, line 252: This looks like it's doing two things at once. One is calculating de/dx at a given energy. Other is producing a full curve.

AtReconstruction/AtFitter/At3DBraggFitter.cxx
```diff
@@ -0,0 +1,354 @@
...
+
+      dEdxVector.push_back(std::make_pair(range / 10., dEdx));
+
+      range += AtFITTER::gDS / 10. / 200;
+   }
+
+   return dEdxVector;
+}
+
+std::vector<std::pair<double, double>> AtFITTER::At3DBraggFitter::GetELossModel(Double_t kineticEnergy)
```
anthoak13, line 280: What is this? How is it different than the previous method?

My brain is staring to get a little tired, so I might have more questions lol

AtReconstruction/AtPatternRecognition/AtSampleConsensus.cxx
```diff
@@ -133,6 +134,14 @@ AtTrack AtSampleConsensus::CreateTrack(AtPattern *pattern, std::vector<const AtH
...
 
    track.SetPattern(pattern->Clone());
+
+   if (auto patternLine = dynamic_cast<AtPatterns::AtPatternLine *>(pattern)) {
+      track.SetGeoTheta(TMath::ATan2(
+         std::abs(patternLine->GetDirection().Z()),
+         TMath::Sqrt(std::pow(patternLine->GetDirection().X(), 2) + std::pow(patternLine->GetDirection().Y(), 2))));
+      track.SetGeoPhi(TMath::ATan2(patternLine->GetDirection().Y(), patternLine->GetDirection().X()));
+   }
+
```
anthoak13, line 144: I think we require c++17 so fine, but we should probably double check our min version isn't 14

AtReconstruction/AtPulseAnalyzer/AtPSA.h
```diff
@@ -49,6 +50,9 @@ class AtPSA {
 
    using HitVector = std::vector<std::unique_ptr<AtHit>>;
 
+   std::uniform_real_distribution<double> fUniform{-0.5, 0.5};
+   std::mt19937_64 fRNG;
```
anthoak13, line 54: I don't like these here. They definitely can't stay as is. They are not inherently thread safe (and PSA methods are called in parallel by parts of the code). Should probably be changed to thread local, and moved to the specific PSA you want them for.

AtReconstruction/AtPulseAnalyzer/AtPSAHitPerTBInRegion.cxx
```diff
@@ -0,0 +1,73 @@
...
+         fIniTB = 0;
+      else
+         fIniTB = limits.first;
+
+      if (limits.second > fNumTbs)
+         fEndTB = fNumTbs;
+      else
+         fEndTB = limits.second;
+   }
+}
```
anthoak13, line 79: This looks like just a better version of the base method. As such, it should just replace the base method - not be overwritten

AtReconstruction/CMakeLists.txt
```diff
@@ -100,12 +102,14 @@ set(SRCS
   AtPatternRecognition/triplclust/src/directedgraph.cxx
   AtPatternRecognition/triplclust/src/postprocess.cxx
   AtPatternRecognition/AtTrackFinderTC.cxx
+  AtPatternRecognition/At3DBraggCurveHandler.cxx
```
anthoak13, line 105: I think something here depends on catima

anthoak13 (changes requested):
The biggest issue that needs fixing is the fact it can't compile without catima!

Other high-level notes:

- [ ] New tab re-implements a lot of AtTabMain. Should either inherit, or just be a normal tab.
- [ ] Catima is tightly connected to the bragg class. Should be integrated as an AtELossModel and that is what is exposed to the Brag classes so we have a consistent handling of energy loss models throughout the code, and to make catima integration useful for things like simulation and other fitters (UKF, MCFitter, etc).
- [ ] I don't understand the purpose of the handler class, it seems necessary?
- [ ] This class should NOT inherit from AtFitter as written. It breaks those conventions
- [ ] We should think about fit results and merge the different fit result data classes. 
- [ ] Data flow (copying tracks, etc. seems fishy. We should thing about that more (related to prev point)

AtData/At3DBraggFitResult.h
RealAurio: I didn't know about that class. I will take a look into it.

AtData/AtTrack.h
RealAurio: Yeah, that makes much more sense. I will probably start with this once I start fixing things.

AtData/AtTrack.h
RealAurio: f3DBraggCurveValues is not the fit result, but rather the experimental values of (ELoss, range) to be fitted. The fit result is in line 51 (fFitResult).

As I initially implemented this, first the At3DBraggCurveHandler class takes the AtTrack and just obtains these (ELoss, range) by projecting the trace of each AtHit into the RANSAC AtPatternLine. Then, after these pairs are obtained, the fitter class is used in order to fit them to a Bragg curve.

My thought in separating this process in these 2 steps is that one can first only use the At3DBraggCurveHandler to check in what order the tail of the Bragg curve is in ADC, so the amplitude factor of the fitter can be tuned, and then repeat the analysis with the fitter on.

Also, since this will eventually be refactored so that it is stored in its own branch, both the (ELoss, range) pairs and the fitting result will be stored in a new AtBraggCurve class object in the AtData folder, maybe?

AtEventDisplay/AtTabs/AtTab3DBraggCurve.h
RealAurio: It's just so that the best fitting dEdx curves for each candidate particle can be calculated and plotted on the top-right window of this tab (I attach a screenshot).

Initially, I was calculating these curves in the At3DBraggFitter class, and then storing them in the At3DBraggFitResult class. However, this made the output files too heavy due to the amount of points needed to have good resolution of the peak, and the analysis was around 4 times slower. So eventually, I just decided to calculate them in the tab with the stored fitted kinetic energy. This makes that panel quite slow and laggy tho, so it probably needs to be improved.

Viewer1

AtEventDisplay/AtTabs/AtTab3DBraggCurve.h
RealAurio: Well, the tab can function without the fitter. If no fitter is provided, then the dEdx curves on the top panel are not plotted. Since right now it's quite laggy, I thought it was a good idea to leave the option to not provide a fitter, which prevents the dEdx curves from being calculated and plotted.

AtEventDisplay/AtTabs/AtTab3DBraggCurve.h
RealAurio: I thought about this, but since in AtTabMain there is a lot of code related to the drawing of the padplane and the traces, which are not needed in this tab, I thought I should create a new one with just the methods I need.

However, it's probably better that it derives from AtTabMain and I just ignore the methods that I do not use, and override the ones that I do use.

AtEventDisplay/AtTabs/AtTab3DBraggCurve.cxx
RealAurio: Agreed, will do.

AtEventDisplay/AtTabs/AtTab3DBraggCurve.cxx
RealAurio: I like the idea of being able to see the AtEvent in the 3D viewer and the Bragg curves in the same tab. In some events, I don't see any peak on the Bragg curve, but by checking in the 3D viewer, I can immediately see that the track reached a wall, so I can discard the event.

Of course, one can just change to the main tab to check the event, but I thought it could be more comfortable if both the histograms and the 3D viewer are in the same tab.

AtEventDisplay/AtTabs/AtTab3DBraggCurve.cxx
RealAurio: Yes, as I replied in a previous comment.

But since the fitting will be refactored so that CATima is used from an AtELossModel class, I can call the AtELossModel from the tab rather than the full fitter.

I just have one concern about this, and it is that I need to input all the information regarding all candidate particles and the material, and with the fitter this was very easy since it was already implemented in there. Maybe, in the At3DBraggFitResult class (or AtBraggCurve once I refactor) I can also store all this information, and the tab can just read it from there.

AtReconstruction/At3DBraggCurveTask.h
RealAurio: The idea was that this task takes the AtTracks from the previous task (ransac), does all the needed calculations regarding the Bragg curve, and stores them inside the AtTrack. Since eventually the fitting results will be stored on their own branch, the output will be just the fitting results in the future.

AtReconstruction/At3DBraggCurveTask.h
RealAurio: As it is right now, I think everything that gets a hold of it.

I'm not sure if it is possible to get to that pointer in any way. The At3DBraggCurveHandler is initialized from the Init of this task, and it is not set externally from a macro with a setter method.

I guess I should make it an unique_ptr just in case?

AtReconstruction/At3DBraggCurveTask.cxx
RealAurio: Yes, I was making a copy of the track object so then I could add the new information regarding the fitting.

Eventually, the fit results will have their own branch, so this will be removed.

AtReconstruction/At3DBraggCurveTask.cxx
RealAurio: SetTrack(AtTrack, AtRawEvent) does not check for null as it is right now, but I guess it makes much more sense to check for it in the SetTrack method rather than in the task. Will change.

AtReconstruction/AtPatternRecognition/At3DBraggCurveHandler.h
RealAurio: Yeah... I actually don't like the name, but I can't come up with anything better.

As I mentioned before, this task only obtains the (ELoss, range) pairs from projecting the AtHits into the AtPatternLine, all the fitting is done afterwards with a fitter class.

I like keeping both steps separated. Right now this class is quite simple, since it only works for the AtPatternLine case. However, in the future I would like to try to make this projection work for an arbitrary trajectory, so maybe the Bragg curve could potentially be obtained for curved tracks.

And also, as I mentioned in a previous comment, I would like the possibility to obtain the (ELoss, range) pairs without making the fit, so that the experimental values can first be examined in order to tune the fit.

AtReconstruction/AtFitter/At3DBraggFitter.h
RealAurio: This is actually the aftermath of an 4-hour long battle between me and this class.

The thing is that Minuit was complaining, since the FCN function needs to be static. After I changed that, FCN was complaining that I was trying to use members of the class. After searching for a solution, I figured that if I made them global, it worked. However, I'm not really happy with this either.

I'm not sure, but maybe if I move them back into the class and make them static, could that work?

AtReconstruction/AtFitter/At3DBraggFitter.h
RealAurio: Actually, Yassid was convinced that I should output AtFittedTracks, but I wasn't sure about that, since I was just fitting the Bragg curve. So, eventually I gave up on the AtFittedTrack. I still kept the At3DBraggFitter as a child of AtFitter, but AtFitter forces me to output AtFittedTracks on that method. However, I do not use that method either.

Probably, I will make it so that At3DBraggFitter does not derive from AtFitter, but I will keep it in the AtFITTER namespace.

AtReconstruction/AtFitter/At3DBraggFitter.h
RealAurio: To comments 1 and 2. Yes, as mentioned in previous replies, I will eventually implement CATima as an AtELossModel.

To comment 3. I was actually trying to keep CATima contained in At3DBraggFitter, and at some point I checked that it could compile without CATima, and it did compile. However, in the last commits I included the At3DBraggFitter inside the AtTab3DBraggCurve and in the At3DBraggCurveTask, so now it tries to compile At3DBraggFitter even if I did not pass CATima. I forgot to make one last check regarding this before creating the pull request... sorry!

Once CATima is contained in an AtELossModel, it should not bleed into the rest of the code.

AtReconstruction/AtFitter/At3DBraggFitter.cxx
RealAurio: I will take a look into this so I can implement it in the future. But for now, minuit stays, since we need something that works for the rcnp experiments.

AtReconstruction/AtFitter/At3DBraggFitter.cxx
RealAurio: This is comparing the range between the one from the ELossModel computed using the GetELossModel method and the experimental ones. iteModel is just there so that for the next bin on the experimental data, it skips all pairs that have already been checked.

AtReconstruction/AtFitter/At3DBraggFitter.cxx
RealAurio: Yes, agreed. I was kind of desperate to get it working xD

As mentioned in a previous reply, I will make it so that it does not inherit from AtFitter anymore, but keep it in the AtFITTER namespace.

AtReconstruction/AtFitter/At3DBraggFitter.cxx
RealAurio: It is actually external. I added this method so it could be called from the tab class, in order to change the particle using the index.

AtReconstruction/AtFitter/At3DBraggFitter.cxx
RealAurio: Yeah, maybe for the estimation of the initial parameter I should compute the curve KineticEnergy = f(range) once for each particle, and then just find the right value in each event.

AtReconstruction/AtFitter/At3DBraggFitter.cxx
RealAurio: I'm not sure if I undertand. Do you mean the GetDeDx and the EstimateEnergyFromRange?

GetDeDx produces a full curve.

EstimateEnergyFromRange computes the estimated kineticEnergy that would produce a track of given range, not the de/dx. This method is only used to make a rough estimate of the kinetic energy so it can be used as initial parameter of the fitter, since I am passing as range for this funtion the x value of the bin with the maximum content.

AtReconstruction/AtFitter/At3DBraggFitter.cxx
RealAurio: Yeah, I should have commented.

The previous function computes the dEdx Vs range curve, with "good resolution". However, this function does essentially the same thing, but integrates dEdx in each bin of the histogram. Since our experimental data is the charge deposited in each bin, we need to compare that with the energy deposited in that bin, which is the integral of dEdx. In the lower panel of the tab, the black curve is the ELoss (integral of dEdx) vs range values obtained with this method, thus the low resolution.

Initially, I only had this method, and then decided to compute the dEdx curves so they could be shown, which is why both methods seem to be doing the same thing. However, I should refactor this so that the GetELossModel calls the GetDeDx, and then integrates de dE/dx bin by bin.

AtReconstruction/AtPatternRecognition/AtSampleConsensus.cxx
RealAurio: I think I don't understand this

AtReconstruction/AtPulseAnalyzer/AtPSA.h
RealAurio: Yeah, I should move them to the AtPSAHitPerTBInRegion. I think initially I had a problem with this, so that's why I moved them here, but I will try to move them back.

AtReconstruction/AtPulseAnalyzer/AtPSAHitPerTBInRegion.cxx
RealAurio: It may be a very noob question but, what's the difference?

AtReconstruction/CMakeLists.txt
RealAurio: At3DBraggCurveHandler doesn't depend on CATima. However, since I included At3DBraggFitter in the At3DBraggCurveTask and in the AtTab3DBraggCurve without thinking, the dependence on CATima is in those classes by accident.

AtReconstruction/AtPulseAnalyzer/AtPSAHitPerTBInRegion.cxx
RealAurio: Ahhhhhhh, I just understood. I am currently redoing everything in this pull request, and I am at this point. I will replace the method in AtPSAHitPerTB directly.
In fact, this whole class only differs on AtPSAHitPerTB in one detail. The base class, after all subHits have been generated, it sets their trace integral to the trace integral of the entire hit, while this subclass sets the trace integral of each subHit to the ADC content of the bin corresponding to it.
I think I will not create a new class, I will simply edit the base one, adding a bool member which controls the setting of the trace integral at the end (set to true by default so nothing that may be using this breaks).

anthoak13:
Replaced by #257

PR #251, opened 2025-06-25
Add first version of unpacker for Si data

anthoak13:


anthoak13:
At the moment appears to be reading all the data correctly, but nothing but FRIB digitizers have data so hard to confirm

PR #254, opened 2025-07-21
Tentative changes for new base class AtFittedTrack.

anthoak13:


AtData/AtFitResult.h
```diff
@@ -0,0 +1,49 @@
...
+ */
+class AtFitResult : public TObject {
+public:
+   using TrackResultPtr = std::unique_ptr<AtFitTrackResult>;
+   using TrackResultsVector = std::vector<TrackResultPtr>;
+   using ResultsMap = std::map<Int_t, TrackResultsVector>;
+
+protected:
+   // Vector to store the results for all different fits done to all tracks in the event.
+   ResultsMap fResults;
```
anthoak13: Needs better comment. What is the map? Like what is the integer? How does it connect to other ints? I don't remember already lol

AtData/AtFittedTrack.h
```diff
@@ -20,125 +23,129 @@ class TClass;
...
+   using TrackResultPtr = std::unique_ptr<AtFitTrackResult>;
+
+   struct Kinematics {
+      Double_t kineticEnergy{-1};    // Kinetic energy
+      Double_t theta{-1};            // Theta scattering angle
+      Double_t phi{-1};              // Phi scattering angle
+      Double_t kineticEnergyXtr{-1}; // Extrapolated kinetic energy
+      Double_t thetaXtr{-1};         // Extrapolated theta scattering angle
+      Double_t phiXtr{-1};           // Extrapolated phi scattering angle
+   };
```
anthoak13: I think this should just be KE, theta, phi. Then we can store two coppies of it. Extr and non-extrap.

anthoak13: Really, this could also just be momentum too. Or some equivalent vector if we don't want to have a new struct?

AtData/AtFittedTrack.h
```diff
@@ -20,125 +23,129 @@ class TClass;
...
+      Double_t chi2{-1};      // Chi2 of the fit
+      Int_t NDF{-1};          // Number of degrees of freedom for the fit
+      Double_t redChi2{-1};   // Reduced chi2
+      Bool_t fitConverged{0}; // Whether or not the fit managed to converge
+   };
+
+   struct ParticleInfo {
+      TString idPDG{""}; // PDG code of the particle
+      Int_t charge{0};   // Charge of the particle
+      Double_t mass{-1}; // Mass of the particle
```
anthoak13: units

AtData/AtFittedTrack.h
```diff
@@ -20,125 +23,129 @@ class TClass;
...
 
-   Float_t fIonChamberEnergy{0};
-   Int_t fIonChamberTime{0};
+   // Vertex where the particle has originated from.
+   std::vector<XYZVector> fVertex;
 
-   Float_t fEnergyXtr{0};
-   Float_t fExcitationEnergyXtr{0};
+   // Track properties.
+   TrackProperties fTrackProperties;
```
anthoak13, line 68: Why is this not a vector but others are vectors?

AtData/AtFittedTrack.h
```diff
@@ -20,125 +23,129 @@ class TClass;
...
 
-   Float_t fDistanceXtr{0};
-   Float_t fTrackLength{0};
-   Float_t fPOCAXtr{0};
+   // Parameters regarding the statistics of the fit.
+   Statistics fStats;
```
anthoak13: This seems unnecessary if we are also storing a AtFitTrackResults which is basically this same information

AtData/AtFittedTrack.h
```diff
@@ -20,125 +23,129 @@ class TClass;
...
-   const std::tuple<Float_t, Float_t, Float_t, Float_t, Float_t, Float_t, Float_t> GetEnergyAngles();
-   const std::tuple<XYZVector, XYZVector, XYZVector> GetVertices();
-   const std::tuple<Float_t, Float_t, Float_t, Float_t, Float_t, Bool_t> GetStats();
-   const std::tuple<Int_t, Float_t, Float_t, Float_t, std::string, Int_t> GetTrackProperties();
-   const std::tuple<Float_t, Float_t> GetIonChamber();
-   const std::tuple<Float_t, Float_t> GetExcitationEnergy();
-   const std::tuple<Float_t, Float_t, Float_t> GetDistances();
+   const Kinematics GetKinematics(int particleIdx) { return fKinematics[particleIdx]; }
+   const ParticleInfo GetParticleInfo(int particleIdx) { return fParticleInfo[particleIdx]; }
+   const XYZVector GetVertex(int particleIdx) { return fVertex[particleIdx]; }
```
anthoak13: set all defaults to =0

AtData/AtFittedTrack.h
```diff
@@ -20,125 +23,129 @@ class TClass;
...
+
+   const Kinematics GetKinematics() { return fKinematics[0]; }
+   const ParticleInfo GetParticleInfo() { return fParticleInfo[0]; }
+   const XYZVector GetVertex() { return fVertex[0]; }
```
anthoak13: remove

AtData/AtFittedTrack.h
```diff
@@ -20,125 +23,129 @@ class TClass;
...
 
    const Int_t GetTrackID() { return fTrackID; }
 
-   const std::tuple<Float_t, Float_t, Float_t, Float_t, Float_t, Float_t, Float_t> GetEnergyAngles();
-   const std::tuple<XYZVector, XYZVector, XYZVector> GetVertices();
-   const std::tuple<Float_t, Float_t, Float_t, Float_t, Float_t, Bool_t> GetStats();
-   const std::tuple<Int_t, Float_t, Float_t, Float_t, std::string, Int_t> GetTrackProperties();
-   const std::tuple<Float_t, Float_t> GetIonChamber();
-   const std::tuple<Float_t, Float_t> GetExcitationEnergy();
-   const std::tuple<Float_t, Float_t, Float_t> GetDistances();
```
anthoak13, line 139: We should keep all of these functions, and mark them depreciated. We do not want to break everyone's analysis codes fully.

AtData/AtFittedTrack.h
```diff
@@ -20,125 +23,129 @@ class TClass;
...
+   TrackResultPtr fTrackResult{nullptr};
 
 public:
    AtFittedTrack() = default;
    ~AtFittedTrack() = default;
 
-   inline void SetTrackID(Int_t trackid) { fTrackID = trackid; }
+   void SetTrackID(Int_t trackid) { fTrackID = trackid; }
 
-   inline void SetEnergyAngles(Float_t energy, Float_t energyxtr, Float_t theta, Float_t phi, Float_t energypra,
```
anthoak13, line 72: Function cannot be removed. Would break analysis code (mark deprecated)

AtData/AtFittedTrack.h
anthoak13: Same with all the setters. It should "look" the same to a user with the old functions. What should be passed as new functions is the structs themselves.

AtData/AtFitTrackResult.h
```diff
@@ -0,0 +1,46 @@
...
+#include <TObject.h>
+
+class TBuffer;
+class TClass;
+class TMemberInspector;
+
+/**
+ * Class for storing the result of the fit of an AtTrack from an AtFitter class.
+ */
+class AtFitTrackResult : public TObject {
```
anthoak13: This name implies there is physics here but it just fit meta-data? Like the stats of the fit?

anthoak13: I think this needs better name or more robust comments describing how everything fits together

AtData/AtTrackingEventOld.h
```diff
@@ -0,0 +1,59 @@
...
+#include <TNamed.h>
+#include <TVector3.h>
+
+#include <vector>
+
+class TBuffer;
+class TClass;
+class TMemberInspector;
+
+class [[deprecated]] AtTrackingEventOld : public AtBaseEvent {
```
anthoak13, line 18: Do we not have a new version of this class?

AtReconstruction/AtFitter/AtFitter.h
```diff
@@ -1,51 +1,71 @@
...
+class AtEvent;
+class AtPatternEvent;
 
 namespace AtFITTER {
 
 class AtFitter : public TObject {
+public:
+   using TrackResultPtr = std::unique_ptr<AtFitTrackResult>;
+   using TrackResultsVector = std::vector<TrackResultPtr>;
+   using TrackResultsSet =
```
anthoak13: Where is this set used?

AtReconstruction/AtFitter/AtFitter.h
```diff
@@ -1,51 +1,71 @@
...
 class TMemberInspector;
-
-namespace genfit {
-class Track;
-} // namespace genfit
+class AtRawEvent;
+class AtEvent;
+class AtPatternEvent;
 
 namespace AtFITTER {
```
anthoak13: I think since we are changing things we should rename this namespace. It doesn't match any others to have AtFITTER with all caps. Maybe just `TrackFit`? or something like that

AtReconstruction/AtFitter/AtFitter.h
```diff
@@ -1,51 +1,71 @@
...
+   using TrackResultsVector = std::vector<TrackResultPtr>;
+   using TrackResultsSet =
+      std::set<TrackResultPtr, std::function<bool(const TrackResultPtr &, const TrackResultPtr &)>>;
+
+protected:
+   // Pointer to the AtPatternEvent to be fitted.
+   AtPatternEvent *fPatternEvent{nullptr};
+
+   // Pointer to the AtFitResult object in which store the fit metadata.
+   AtFitResult *fFitResult{nullptr};
```
anthoak13: Why is this a raw pointer? This is like an output no? So who owns this object?

AtReconstruction/AtFitter/AtFitter.h
```diff
@@ -1,51 +1,71 @@
...
-namespace genfit {
-class Track;
-} // namespace genfit
+class AtRawEvent;
+class AtEvent;
+class AtPatternEvent;
 
 namespace AtFITTER {
 
 class AtFitter : public TObject {
```
anthoak13, line 30: Something is weird about the interface of this class. I can't quite put my finger on what I don't like about it though but it is probably worth talking about. I think I no longer have it in my head how all the pieces are supposed to fit together

AtReconstruction/AtFitter/AtFitter.h
```diff
@@ -1,51 +1,71 @@
...
+   virtual bool CompareTrackFitsFunction(const TrackResultPtr &trackResultA, const TrackResultPtr &trackResultB) = 0;
 
 public:
-   AtFitter();
-   virtual ~AtFitter();
-   virtual std::vector<std::unique_ptr<AtFittedTrack>> ProcessTracks(std::vector<AtTrack> &tracks) = 0;
+   AtFitter() = default;
+   ~AtFitter() = default;
+
+   virtual std::vector<std::unique_ptr<AtFittedTrack>> ProcessEvent() = 0;
```
anthoak13: I don't think this is the right return type. Or maybe it is. It depends on how this class is supposed to work. If it's supposed to be fitting an AtPatternEvent then the return type should be an AtTrackingEvent. If it's supposed to be acting on an AtTrack then the return type should be an AtFittedTrack or something.

AtReconstruction/AtFitter/AtFitter.h
```diff
@@ -1,51 +1,71 @@
...
 namespace AtFITTER {
 
 class AtFitter : public TObject {
+public:
+   using TrackResultPtr = std::unique_ptr<AtFitTrackResult>;
+   using TrackResultsVector = std::vector<TrackResultPtr>;
+   using TrackResultsSet =
+      std::set<TrackResultPtr, std::function<bool(const TrackResultPtr &, const TrackResultPtr &)>>;
+
+protected:
```
anthoak13: I don't think it makes much sense to store temporary pointers that users have to remember to manage

AtReconstruction/AtFitterTask.h
```diff
@@ -41,21 +40,22 @@ class AtTrackTransformer;
...
+   ~AtFitterTask() = default;
 
    void SetInputBranch(TString branchName);
    void SetOutputBranch(TString branchName);
    void SetPersistence(Bool_t value = kTRUE);
 
+   void SetRawEventBranch(TString branchName);
+   void SetEventBranch(TString branchName);
+
+   void SetFitResultBranch(TString branchName);
```
anthoak13: So is AtFitResult actually the equivalent of AtFittedTrack?

AtReconstruction/AtFitterTask.h
```diff
@@ -41,21 +40,22 @@ class AtTrackTransformer;
 namespace AtFITTER {
 class AtFitter;
 } // namespace AtFITTER
-namespace genfit {
-class Track;
-} // namespace genfit
 
 class AtFitterTask : public FairTask {
```
anthoak13, line 50: Probably needs a very robust doc string about the logic.

AtReconstruction/AtFitterTask.cxx
```diff
@@ -84,22 +112,39 @@ void AtFitterTask::SetParContainers()
...
+      fFitter->SetEvent(event);
+   }
+
    fTrackingEventArray.Delete();
 
    auto trackingEvent = dynamic_cast<AtTrackingEvent *>(fTrackingEventArray.ConstructedAt(0));
+   auto fitResult = dynamic_cast<AtFitResult *>(fFitResultArray.ConstructedAt(0));
 
-   std::cout << " Event Counter " << fEventCnt << "\n";
+   LOG(info) << " AtFitterTask::Exec() : Fitting event " << fEventCnt;
```
anthoak13: Remove hard coded function name

AtReconstruction/AtFitterTask.cxx
```diff
@@ -84,22 +112,39 @@ void AtFitterTask::SetParContainers()
...
 
-   AtPatternEvent &patternEvent = *(dynamic_cast<AtPatternEvent *>(fPatternEventArray->At(0)));
-   std::vector<AtTrack> &tracks = patternEvent.GetTrackCand();
-   std::cout << " AtFitterTask:Exec -  Number of candidate tracks : " << tracks.size() << "\n";
+   AtPatternEvent *patternEvent = dynamic_cast<AtPatternEvent *>(fPatternEventArray->At(0));
+   std::vector<AtTrack> &tracks = patternEvent->GetTrackCand();
+   LOG(info) << " AtFitterTask::Exec() : Number of candidate tracks : " << tracks.size();
```
anthoak13: remove func name

AtData/AtFitResult.h
RealAurio: You are right, the comment could be better.

The integer in this map connects to the fTrackId for which the fits were performed, while the vector of AtFitTrackResult is the collection of fit results of all fits performed for the same track (for example, take one track, assume it could be either proton, deuterium or tritium, and check all 3 cases). Will update this comment.

AtData/AtFitTrackResult.h
RealAurio: I think we wanted to name this AtFitMetadata, but for some reason I forgot about this. I can change the naming of the class. However, the now derived class AtMCResult has result in the naming, but I'm afraid of changing the naming in AtMCFitResult since it was an already existing class and would imply changing the convention in a lot of places.

For now, I will change the name of the base class and leave AtMCResult unchanged, but please let me know if you still want to change the naming of AtMCResult.

AtData/AtFittedTrack.h
RealAurio: In the end, we are interested in getting kinetic energy and angles, but I could also store a TLorentzVector.

Also, I initially saved 2 copies of Kinematics, one for extr and one for non-extr. I guess I will change it back to that.

AtData/AtFittedTrack.h
RealAurio: Agreed.

AtData/AtFittedTrack.h
RealAurio: Others are vectors because we are considering that 1 AtTrack could contain multiple particles, thus multiple entries in the vectors. However, for track properties, these are properties of the track as a whole, so it should only be 1 entry.

Please check the definition of TrackProperties just in case you see a variable that should not live in there based on this.

AtData/AtFittedTrack.h
RealAurio: Yes, this was here before we had the idea to store the AtTRackFitResult in here too. Since we are now storing it, these statistics are no longer needed to be stored separately. I will remove this now.

AtData/AtFittedTrack.h
RealAurio: Okay.

AtData/AtFittedTrack.h
RealAurio: Agreed, will bring these templates back and change the implementation so it adapts to the refactored class.

AtData/AtFittedTrack.h
RealAurio: Okay.

AtData/AtFittedTrack.h
RealAurio: Should I remove lines 143 and 144 too?

AtData/AtTrackingEventOld.h
RealAurio: I think the new version did in fact not need any new changes, so it remained unchanged and thus it does not appear in this commit (the new AtFittedTrack is still named AtFittedTrack). As for the old one, I had to change AtFittedTrack to AtFittedTrackOld, and stuff like that, so this new class was needed with just that little change.

RealAurio (commented):
First changes I will work on. I added a couple of questions as to how I should approach this.

AtData/AtFitTrackResult.h
RealAurio: Renaming from result to metadata done in next 2 commits.

AtData/AtFitResult.h
RealAurio: Done in 3rd commit after this one.

AtData/AtFittedTrack.h
RealAurio: Done in 3rd commit after this one.

AtData/AtFittedTrack.h
RealAurio: Done in 3rd commit after this one.

AtData/AtFittedTrack.h
RealAurio: Line 144 was removed by removing the statistics.

RealAurio (commented):
Some changes

AtData/AtFittedTrack.h
RealAurio: Actually, there is a problem with making it look the same as with the old functions, since we removed the IC from this class for example, we could not keep the getter or setter of the IC as it was before. In cases like these, I will keep the function, marked as deprecated, and it will simply print that the IC no longer exists in this class.

AtData/AtFittedTrack.h
RealAurio: Separated the kinematics in 2 different members, one for non-extrap and one for extrap. This is in 4th commit after this one

AtData/AtFittedTrack.h
RealAurio: Or I bring back removed members that have not been moved into a new struct (simply removed), and mark them as deprecated.

AtData/AtFittedTrack.h
RealAurio: Done in 4th commit after this one.

AtData/AtFittedTrack.h
RealAurio: Done in 4th commit after this one.

AtData/AtFittedTrack.h
RealAurio: Done in 4th commit after this one.

AtData/AtFittedTrack.h
RealAurio: Ah, just realized, this is related to setting defaults to 0. Done in 4th commit after this one.

AtReconstruction/AtFitter/AtFitter.h
RealAurio: Probably something similar as to what you suggested in the Bragg curve pull request #257. I will work on this to implement a similar interface.

AtReconstruction/AtFitter/AtFitter.h
RealAurio: Will be used internally by the fitter. The ordering is meant to put the "best" fit first, which is done with a set with a comparison function. This criteria for "best" fit may change depending on the specific fitter that was used.

AtReconstruction/AtFitter/AtFitter.h
RealAurio: Yes, as in #257. Will work on this.

AtReconstruction/AtFitter/AtFitter.h
RealAurio: This AtFitMetadata (previously AtFitResult) is created in the AtFitterTask, in a separate branch. Then, the pointer is passed to the AtFitter so that it can be modified accordingly. Should I use other type of pointer?

AtReconstruction/AtFitter/AtFitter.h
RealAurio: I left the return type as it was before. But maybe it makes sense that it returns an AtTrackingEvent, so we can follow a similar interface as in #257.

AtReconstruction/AtFitterTask.h
RealAurio: No, it was a bad choice of name. It should be FitMetadata. This was already changed to the new naming.

AtReconstruction/AtFitter/AtFitter.h
RealAurio: TrackFit would imply that the fitters always act on the AtTracks. However, if I understood correctly, for the case of the MCFitter, the fitting is done on the event as a whole, not track by track.

I will change the name to something like EventFit, so it's a bit more general. What do you think?

RealAurio: I'm pretty sure this would slightly break someones code though, but I also believe that is going to be inevitable if we want to do this refactor.

AtReconstruction/AtFitter/AtFitter.h
RealAurio: I'm making all current fitters derive from AtFitterOld, since changing the interface would break those. At some point we could change them too so they derive from this new one, but I don't this is a priority right now as long as they are still functional.

RealAurio: Well only AtGenfit actually derives from AtFitter, so that's the only one I need to change to AtFitterOld. I mark it as deprecated and put a message as a reminder to change it eventually.

AtReconstruction/AtFitterTask.cxx
RealAurio: Done in 5th commit after this one

AtReconstruction/AtFitterTask.cxx
RealAurio: Done in 5th commit after this one

AtReconstruction/AtFitter/AtFitter.h
RealAurio: Done in 5th commit after this one

AtReconstruction/AtFitter/AtFitter.h
RealAurio: Done in 5th commit after this one.

AtReconstruction/AtFitter/AtFitter.h
RealAurio: These pointers were removed in 5th commit after this one anyways.

AtReconstruction/AtFitter/AtFitter.h
RealAurio: Done in 5th commit after this one.

AtReconstruction/AtFitter/AtFitter.h
RealAurio: Done in 5th commit after this one.

AtReconstruction/AtFitterTask.h
RealAurio: Done in 6th commit after this one.

AtData/AtFittedTrack.h
anthoak13: I think most of the track properties seem to be particle-track related rather than AtTrack related. Like the dedx would not make sense to talk about an entire event. Looking through the list I think I would say all of those parameters are probably particle specific so I would put this as a vector.

AtData/AtFittedTrack.h
anthoak13: I am happy to keep it this way. Or for functions that partially set the removed member variables we mark them deprecated. Functions that only set removed members are removed?

I'm not sure what the best option is. I think we will break things regardless, but the less the better I suppose. Maybe we try to run some old codes to see how much a pain these different changes are to deal with?

AtReconstruction/AtFitter/AtFitter.cxx
```diff
@@ -1,268 +1,22 @@
...
-
-AtFITTER::AtFitter::AtFitter() = default;
-
-AtFITTER::AtFitter::~AtFitter() = default;
-
-std::tuple<Double_t, Double_t> AtFITTER::AtFitter::GetMomFromBrho(Double_t M, Double_t Z, Double_t brho)
+void EventFit::AtFitter::FitEvent(AtTrackingEvent *trackingEvent, AtPatternEvent *patternEvent,
+                                  AtFitMetadata *fitMetadata, AtRawEvent *rawEvent, AtEvent *event)
 {
+   // Extract the candidate AtTracks.
```
anthoak13: Should we add checks for null pointer here?

anthoak13 (commented):
I think this looks good? Is there anything else you would want to change before we try and merge?

AtData/AtFittedTrack.h
RealAurio: Functions that only set removed members are not removed in the end. You can see in this commit that SetIonChamber and SetExcitationEnergy are back. They are marked as deprecated, and also the fIonChamberEnergy, fIonChamberTime, fExcitationEnergy and fExcitationEnergyXtr that they used to set, those are also brought back, but marked as deprecated and advised to not be used.

Then, for cases where the old quantities are now saved in one of the structs, the old setter sets that quantity inside the struct.

Also, if you check the setters, the mapping between setter->member/member of struct->getter is done so that what you set is still what you get under the old convention, so in principle that should still work, with a bunch of deprecation warnings that is.

AtData/AtFittedTrack.h
RealAurio: Well, it depends. Maybe dEdx is not the best example, since it should be particle specific. But for example, the total charge is meant to be the total charge of the track, being it 1, 2 or 3 particles, same with number of hits, it's meant to be total hits in all the track. The extrapolated distance is meant to be the total distance that needed to be extrapolated from the first hit to the real vertex along the axis of the  ATTPC, for the part of the track that comes out of the beam region in case we are talking about the fission pattern.

Meanwhile, real particle vertex would be stored in  the fVertices vector, one for each particle.

The ony one that maybe feel out of place to me is the dEdx, but in a case where there is only 1 particle in the track, it can still make sense, and maybe it can simply be ignored in other cases. Maybe track lengths also do not make sense in case that there is more than 1 particle.

AtReconstruction/AtFitter/AtFitter.cxx
RealAurio: This would be something similar as the last change that you asked for in #257. I can add a quick check. If no pattern event, then return.

RealAurio: Also for tracking event actually.

Raw event, event and fit metadata can be nullptr considering how GetFittedTrack is defined.

RealAurio: Done in last commit.

anthoak13:
@RealAurio are there any other changes you want to make to this before merging? I've been slowly working on trying to implement the KF as a task under the refactor so it would be useful to merge this before the LF tal gets merger.

RealAurio:
> @RealAurio are there any other changes you want to make to this before merging? I've been slowly working on trying to implement the KF as a task under the refactor so it would be useful to merge this before the LF tal gets merger.

@anthoak13  Oh, sorry, I did not see this message until now. I think I do not want to add anything else here. It has been a while since we last looked at this, so maybe it is worth giving it one last review.

Actually, there are a lot of deprecation warnings on compilation due to how we kept the old classes and methods. I do not know if it is possible to fix that somehow.

PR #257, opened 2025-08-04
Implementation of the algorithm to find the experimental Bragg curves from the AtTracks.

RealAurio:
Added container in AtTrack in order to store the experimental values of (range, eLoss).

Added a base class for possible classes that modify a given AtPatternEvent.

Added the task to apply AtPatternModifications to the AtPatternEvents.

Added the option to pass AtEvent and AtRawEvent to the AtPatternModification.

Added a virtual method to the base class AtPattern that computes the distance between 2 points along the pattern. Implemented it for the case of AtPatternLine.

Implemented a virtual method in base class AtPattern to obtain the parameter at a given point of the AtPattern.

Modified the AtPSAHitPerTB.

Implemented an AtPatternModification which extracts the pair of values for the Bragg curve reconstruction and saves them in the AtTracks.

Added quick test macros to check if this feature works so far.

Generated the Bragg curve histogram and saved it in a new BraggCurve struct in AtTrack.

Added an AtTab to visualize the Bragg curves of the AtTacks.

Implemented the smoothing of the Bragg curves.


AtData/AtPattern/AtPattern.h
```diff
@@ -150,6 +150,20 @@ class AtPattern : public TObject {
...
+    * This function needs to be overriden for any specific AtPattern subclass in order to be useful.
+    */
+   virtual Double_t DistanceAlongPattern(double t1, double t2) const { return -99999; }
+
+   /**
+    * Parameter value at the point passed as input.
+    * By default returns -99999, which should not be possible in any case.
+    * This function needs to be overriden for any specific AtPattern subclass in order to be useful.
+    */
+   virtual double parameterAtPoint(const XYZPoint &point) const { return -99999; }
```
anthoak13: Should be capital P following convention. I would also move it to line 102ish where the related functions are.,

AtData/AtPattern/AtPattern.h
```diff
@@ -150,6 +150,20 @@ class AtPattern : public TObject {
    void SetPatternPar(std::vector<double> par) { fPatternPar = std::move(par); }
    void SetChi2(double chi2) { fChi2 = chi2; }
 
+   /**
+    * Calculate the distance in mm along the line from point at parameter t1 to point at parameter t2.
+    * By default returns -99999, which should not be possible in any case.
+    * This function needs to be overriden for any specific AtPattern subclass in order to be useful.
+    */
+   virtual Double_t DistanceAlongPattern(double t1, double t2) const { return -99999; }
```
anthoak13: By default this should probably just be something like:
```cpp
return (GetPointAt(t1) - GetPointAt(t2)).R();
```

AtData/AtPattern/AtPatternLine.cxx
```diff
@@ -210,3 +210,12 @@ std::vector<Double_t> AtPatternLine::lineIntersecR(Double_t rMax, Double_t tMin,
...
    return result;
 }
+
+Double_t AtPatternLine::DistanceAlongPattern(double t1, double t2) const
+{
+   XYZPoint point1 = GetPointAt(t1);
+   XYZPoint point2 = GetPointAt(t2);
+   Double_t distanceSquared = std::pow(point1.X() - point2.X(), 2) + std::pow(point1.Y() - point2.Y(), 2) +
+                              std::pow(point1.Z() - point2.Z(), 2);
+   return TMath::Sqrt(distanceSquared);
```
anthoak13: Doesn't need to change, but equivalent to `(point1-point2).R()` which is maybe easier to read

AtData/AtPattern/AtPatternLine.h
```diff
@@ -43,11 +43,13 @@ class AtPatternLine : public AtPattern {
 
    TEveLine *GetEveLine(Double_t rMax = 250) const;
 
+   virtual Double_t DistanceAlongPattern(double t1, double t2) const override;
+   double parameterAtPoint(const XYZPoint &point) const override;
```
anthoak13: I didn't see this implementation in the cxx?

AtData/AtPattern/AtPattern.h
anthoak13: I'm not sure I love exposing this internal parameter? How the pattern is defined, it might change randomly since it is an "internal" parameter. It would I think make more sense to base it on "closest point".

So `DistanceAlongPattern` takes in (XYZPos point1, XYZPos point2), rather than the parameter. That way the function behavior is more fixed. If we change the internal parameterization, it won't break the code.

AtData/AtTrack.h
```diff
@@ -24,6 +24,16 @@ class TMemberInspector;
...
+public:
+   struct BraggCurve {
+      std::vector<Double_t> IntegratedELossValues;
+      std::vector<Double_t> RangeValues;
+      std::vector<Double_t> ELossErrors;
+      Int_t nBins{0};
+      Double_t binSize{0};
+      Int_t smoothingSteps{0};
+   };
+
```
anthoak13, line 36: I think this should not live in AtTrack but something like `AtTrackBragg` which inherits from AtTrack.

AtData/AtTrack.h
```diff
@@ -44,6 +54,12 @@ class AtTrack : public TObject {
    std::pair<Double_t, Double_t> fGeoCenter; //< Center of the spiral track
    std::vector<AtHitCluster> fHitClusterArray; //< Clusterized hits container. Can also be stored in fHitArray
 
+   // Obtained in AtPatternModification
+   // Vector of pair of values for the Bragg curve of the track (archLength[mm], eLoss[a.u.]).
+   std::vector<std::pair<Double_t, Double_t>> fBraggCurveValues;
+   // Container for the Bragg curve integrated over the binning.
+   BraggCurve fBraggCurve;
```
anthoak13, line 61: Then these exist only in `AtTrackBragg` too.

AtData/AtTrack.h
```diff
@@ -89,6 +108,9 @@ class AtTrack : public TObject {
    std::pair<Double_t, Double_t> GetGeoCenter() const { return fGeoCenter; }
    std::vector<AtHitCluster> *GetHitClusterArray() { return &fHitClusterArray; }
 
+   std::vector<std::pair<Double_t, Double_t>> GetBraggCurveValues() { return fBraggCurveValues; }
+   const BraggCurve GetBraggCurve() { return fBraggCurve; }
```
anthoak13: Both functions should be marked `const` since their behavior is actually const

anthoak13: The return type being marked `cosnt` is unnecessary. You're returning a copy, to there is no reason to mark it const.

AtEventDisplay/AtTabs/AtTabBraggCurve.cxx
```diff
@@ -0,0 +1,169 @@
...
+namespace DataHandling {
+class AtSubject;
+}
+
+ClassImp(AtTabBraggCurve);
+
+AtTabBraggCurve::AtTabBraggCurve() : AtTabMain() {}
+
+AtTabBraggCurve::~AtTabBraggCurve()
+{
```
anthoak13, line 36: You should call the parent deconstructor and only handle your new additions.

AtEventDisplay/AtTabs/AtTabBraggCurve.h
```diff
@@ -0,0 +1,49 @@
...
+class AtTabBraggCurve : public AtTabMain {
+protected:
+   TCanvas *fCvsELossVRange{nullptr};
+   TH1F *fHistELossVRange{nullptr};
+
+   int fTrackIdx{-1};
+
+public:
+   AtTabBraggCurve();
+   ~AtTabBraggCurve();
```
anthoak13: Should be marked `virtual`

AtEventDisplay/AtTabs/AtTabBraggCurve.cxx
anthoak13: As written you are freeing resources twice. When this object is deconstructed, all base deconstructors are also called in reverse order of construction.

anthoak13: So the only thing your deconstructor here should manage is 
```
 TCanvas *fCvsELossVRange{nullptr};
 TH1F *fHistELossVRange{nullptr};\
```

AtEventDisplay/AtTabs/AtTabBraggCurve.h
```diff
@@ -0,0 +1,49 @@
...
+   virtual void Update(DataHandling::AtSubject *sub) override;
+
+protected:
+   virtual void MakeTab(TEveWindowSlot *slot) override;
+
+   void DrawHistELossVRange();
+   void DrawHistELossVRange(AtTrack::BraggCurve braggCurve);
+
+private:
+   virtual void UpdatePatternEventElements() override;
```
anthoak13: Should probably be protected if override. Or marked as final.

AtReconstruction/AtPatternModification/AtPatternModification.h
```diff
@@ -0,0 +1,38 @@
+#ifndef AtPATTERNMODIFICATION_H
+#define AtPATTERNMODIFICATION_H
+
+#include "AtEvent.h"
+#include "AtPatternEvent.h"
+#include "AtRawEvent.h"
+
+class AtPatternModification : public TObject {
```
anthoak13: Should not inherit from TObject. This is never written to disk.

AtReconstruction/AtPatternModification/AtPatternModification.h
```diff
@@ -0,0 +1,38 @@
...
+
+   // Mandatory to set.
+   void SetPatternEvent(AtPatternEvent *patternEvent) { fPatternEvent = patternEvent; }
+
+   // Optional to set.
+   void SetRawEvent(AtRawEvent *rawEvent) { fRawEvent = rawEvent; }
+   void SetEvent(AtEvent *event) { fEvent = event; }
+
+   // Reset pointers.
+   void Reset();
```
anthoak13: I think this interface is overly cumbersome. It requires the user to set and reset various pointers, when that seems like it is unnecessary. I think this entire interface should be:

```cpp
#ifndef AtPATTERNMODIFICATION_H
#define AtPATTERNMODIFICATION_H

class AtEvent.h;
class AtPatternEvent.h;
class AtRawEvent.h;

class AtPatternModification {
virtual void Init() = 0;
/**
* Modifies patternEvent.
*/
virtual void ModifyPatternEvent(AtPatternEvent *patternEvent, const AtRawEvent *raw=nullptr, const AtEvent* event = nullptr) = 0;

};

File: AtReconstruction/AtPatternModification/AtPatternModification.h [hunk 13]
```diff
@@ -0,0 +1,38 @@
...
+
+private:
+   ClassDef(AtPatternModification, 1);
```
anthoak13: No class def

AtReconstruction/AtReconstructionLinkDef.h
```diff
@@ -37,6 +37,10 @@
 #pragma link C++ class SampleConsensus::AtEstimator - !;
 #pragma link C++ enum SampleConsensus::Estimators;
 
+#pragma link C++ class AtPatternModification + ;
+#pragma link C++ class AtBraggCurveFinder + ;
+#pragma link C++ class AtPatternModificationTask + ;
```
anthoak13: All of these should be -!, not + I think. None of them are data structures so be written to disk.

AtReconstruction/AtPatternModificationTask.h
```diff
@@ -0,0 +1,49 @@
...
+
+   std::vector<std::unique_ptr<AtPatternModification>> fPatternModifications;
+   TClonesArray *fPatternEventArray;
+   TClonesArray *fRawEventArray;
+   TClonesArray *fEventArray;
+
+   ULong_t fEventCnt{0};
+
+public:
+   AtPatternModificationTask();
```
anthoak13: Constructor should take the pattern modification class.

AtReconstruction/AtPatternModificationTask.h
```diff
@@ -0,0 +1,49 @@
...
+   ~AtPatternModificationTask() = default;
+
+   void SetPatternEventBranch(TString branchName);
+   void SetRawEventBranch(TString branchName);
+   void SetEventBranch(TString branchName);
+
+   void AddPatternModification(std::unique_ptr<AtPatternModification> patternModification)
+   {
+      fPatternModifications.push_back(std::move(patternModification));
+   }
```
anthoak13: Remove

AtReconstruction/AtPatternModificationTask.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+class AtPatternModificationTask : public FairTask {
+private:
+   TString fPatternEventBranchName;
+   TString fRawEventBranchName;
+   TString fEventBranchName;
```
anthoak13: Give defaults for PatternEventBranch

AtReconstruction/AtPatternModificationTask.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <Rtypes.h>
+#include <TClonesArray.h>
+#include <TString.h>
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+class AtPatternModificationTask : public FairTask {
```
anthoak13, line 22: I think this needs to not do in-place modification. Or at least that should be an option and it only does in-place modification of the output and input branches are the same

AtReconstruction/AtPatternModification/AtPatternModification.h
```diff
@@ -0,0 +1,38 @@
...
+#include "AtRawEvent.h"
+
+class AtPatternModification : public TObject {
+protected:
+   // Pointer to the AtPatternEvent to be modified.
+   AtPatternEvent *fPatternEvent{nullptr};
+
+   // Pointers to AtRawEvent and AtEvent that may be used or not in the modification.
+   AtRawEvent *fRawEvent{nullptr};
+   AtEvent *fEvent{nullptr};
```
anthoak13: Remove

AtReconstruction/AtPatternModification/AtPatternModification.h
anthoak13: Remove

AtReconstruction/AtPatternModification/AtPatternModification.h
```diff
@@ -0,0 +1,38 @@
...
+
+public:
+   AtPatternModification() = default;
+   ~AtPatternModification() = default;
+
+   virtual void ModifyPatternEvent() = 0;
```
anthoak13: Take in pointers explicitly

AtReconstruction/AtPatternModificationTask.h
```diff
@@ -0,0 +1,49 @@
...
+   TClonesArray *fRawEventArray;
+   TClonesArray *fEventArray;
+
+   ULong_t fEventCnt{0};
+
+public:
+   AtPatternModificationTask();
+   ~AtPatternModificationTask() = default;
+
+   void SetPatternEventBranch(TString branchName);
```
anthoak13: Set input and output branches

AtReconstruction/AtPatternModificationTask.cxx
```diff
@@ -0,0 +1,84 @@
...
+   fRawEventArray = dynamic_cast<TClonesArray *>(ioMan->GetObject(fRawEventBranchName));
+   if (fRawEventArray == nullptr) {
+      LOG(info) << "AtRawEvent branch name was not set. No AtRawEvent will be passed to the AtPatternModifications.";
+   }
+
+   fEventArray = dynamic_cast<TClonesArray *>(ioMan->GetObject(fEventBranchName));
+   if (fEventArray == nullptr) {
+      LOG(info) << "AtEvent branch name was not set. No AtEvent will be passed to the AtPatternModifications.";
+   }
+
```
anthoak13, line 67: Needs to check if output is same as input. If not, register output branch

AtReconstruction/AtPatternModificationTask.cxx
```diff
@@ -0,0 +1,84 @@
...
+   return kSUCCESS;
+}
+
+void AtPatternModificationTask::Exec(Option_t *option)
+{
+   if (fPatternEventArray->GetEntriesFast() == 0)
+      return;
+
+   LOG(info) << " AtPatternModificationTask::Exec() : Applying pattern modifications to pattern event " << fEventCnt;
```
anthoak13: Don't add "AtPatternModificationTask::Exec() :" to logs. Use FairLogger verbosity in your macros enable that if you want, but it should not be forced on. (https://github.com/FairRootGroup/FairLogger?tab=readme-ov-file#4-verbosity)

AtReconstruction/AtPatternModificationTask.cxx
```diff
@@ -0,0 +1,84 @@
...
+
+   AtPatternEvent *patternEvent = dynamic_cast<AtPatternEvent *>(fPatternEventArray->At(0));
```
anthoak13, line 80: If input and output not the same, need to construct the output pattern event through copy constructor.

AtReconstruction/AtPatternModification/AtPatternModification.h
anthoak13: Thinking about this some more. The natural unit to act on is an AtTrack, so that should be the primary hook of this interface. So I think it should actually looks like:

```cpp
#ifndef AtPATTERNMODIFICATION_H
#define AtPATTERNMODIFICATION_H

class AtEvent.h;
class AtPatternEvent.h;
class AtRawEvent.h;

class AtPatternModification {
virtual void Init() = 0;
/**
* Modifies patternEvent.
*/
virtual void ModifyPatternEvent(AtPatternEvent *patternEvent, const AtRawEvent *raw=nullptr, const AtEvent* event = nullptr);
virtual void ModifyTrack(AtTrack *track, const AtRawEvent *raw=nullptr, const AtEvent* event = nullptr) = 0; /// Modifies track in place

};
```

AtReconstruction/AtPatternModification/AtPatternModification.cxx
```diff
@@ -0,0 +1,10 @@
+#include "AtPatternModification.h"
+
+ClassImp(AtPatternModification);
+
+void AtPatternModification::Reset()
+{
+   fPatternEvent = nullptr;
+   fRawEvent = nullptr;
+   fEvent = nullptr;
+}
```
anthoak13: Header just implements a default version of `ModifyPatternEvent` which loops through and calls ModifyTrack on each track in the event.

AtReconstruction/AtPatternModification/AtBraggCurveFinder.cxx
```diff
@@ -0,0 +1,162 @@
...
+   std::vector<AtTrack> &tracks = fPatternEvent->GetTrackCand();
+   LOG(info) << "Applying AtBraggCurveFinder modification to " << tracks.size() << " tracks.";
+
+   if (fRawEvent == nullptr)
+      LOG(warning) << "No AtRawEvent was passed to the At3DBraggFinder, so it will not be used. The resulting bragg "
+                      "curve may have more fluctuations in the ELoss per bin than expected!";
+
+   for (auto &track : tracks)
+      ProcessTrack(track);
+}
```
anthoak13: This is then just the base class implementation so doesn't need to be overwritten in your case.

AtReconstruction/AtPatternModification/AtBraggCurveFinder.cxx
```diff
@@ -0,0 +1,162 @@
...
+
+   track.AddBraggCurvePair(range, eLoss);
+}
+
+void AtBraggCurveFinder::GenerateBraggCurveHistogram(AtTrack &track)
+{
+
+   // Define the histogram where to integrate the charge over bins.
+   int nBins = std::ceil(fMaxLength / fBinSize);
+   TH1F *histBraggCurve = new TH1F("histBraggCurve", "histBraggCurve", nBins, 0, nBins * fBinSize);
```
anthoak13, line 117: This is a memory leak I am pretty sure., The memory is never cleaned up

AtReconstruction/AtPulseAnalyzer/AtPSAHitPerTB.cxx
```diff
@@ -24,7 +24,7 @@ AtPSAHitPerTB::HitVector AtPSAHitPerTB::AnalyzePad(AtPad *pad)
 {
    auto pos = pad->GetPadCoord();
    if ((pos.X() < -9000 || pos.Y() < -9000) && pad->GetPadNum() != -1)
-      LOG(error) << " AtPSAHitPErTB::Analysis Warning! Wrong Coordinates for Pad : " << pad->GetPadNum() << std::endl;
+      LOG(error) << " AtPSAHitPerTB::Analysis Warning! Wrong Coordinates for Pad : " << pad->GetPadNum() << std::endl;
```
anthoak13: Hardcoding what should be a verbosity setting

anthoak13 (changes requested):
I think we want to tweak the interface for pattern modification and the associated task

AtData/AtPattern/AtPatternLine.h
RealAurio: parameterAtPoint was already implemented in the AtPatternLine, I just moved it from protected to public so that it can be used in the AtBraggCurveFinder for example.

However, there might be a workaround, but I am not sure if it would work for every pattern.

AtData/AtPattern/AtPattern.h
RealAurio: This would only make sense for a line for example. For a curved pattern, we would need to integrate the archlength over the curved trajectory that defines the pattern. Maybe I can leave this as default, but it would need to be overriden in other patterns in any case.

AtData/AtPattern/AtPattern.h
RealAurio: It was without capital P since it was implemented already in AtPatternLine. I did not want to change it to capital P just in case I would break anything.

Also, I need this parameter so that I can compute the distance between 2 points.

There might be a workaround tho, I will think about this.

AtData/AtPattern/AtPatternLine.cxx
RealAurio: Yeah, thanks for this, I was getting annoyed while writing it...

AtData/AtTrack.h
RealAurio: Yeah, maybe makes sense. AtPatternEvent should still work with these I guess.

AtReconstruction/AtPatternModification/AtPatternModification.h
RealAurio: The AtRawEvent and AtEvent? I do need the AtRawEvent in order to construct the Bragg curves, and thought that may leave the AtEvent there just in case some future modification may need it.

RealAurio: I just saw comments below, will do

AtReconstruction/AtPatternModificationTask.h
RealAurio: It should take a vector of them, since multiple modifications may be applied sequentally on the same task.

AtData/AtPattern/AtPattern.h
RealAurio: Done in next commit

AtData/AtPattern/AtPattern.h
RealAurio: Removed from base class in next commit.

AtData/AtPattern/AtPatternLine.cxx
RealAurio: Done in next commit.

AtData/AtPattern/AtPatternLine.h
RealAurio: Moved it back to protected. Left the non-capital p since it was like that before (old line 50) and I don't want to break something due to this.

AtData/AtTrack.h
RealAurio: Tried to do this, but there is a major problem. The AtTracks are stored inside AtPatternEvent as a vector of AtTracks, not as a vector of pointers to AtTracks. This makes it impossible to store objects of a derived class from AtTrack, since the added methods and attributes would be sliced off.

I tried to modify AtPatternEvent so that it stores unique pointers of AtTRacks instead, in the same style of any other container class in ATTPCROOTv2. As expected, this raises many errors in infinite parts of the code.

Therefore, even if this is a good idea, I would first refactor AtPatternEvent in a separate feature, since it doesn't make much sense to implement such a big change in a bragg curve feature. After that refactoring is implemented, I would then come back to this and change it so that the bragg curve information is stored inside a new derived class of AtTrack.

AtReconstruction/AtPatternModification/AtPatternModification.h
RealAurio: Changed in second commit after this one.

AtReconstruction/AtPatternModification/AtPatternModification.h
RealAurio: Changed in second commit after this one.

AtReconstruction/AtPatternModification/AtPatternModification.h
RealAurio: Changed in second commit after this one.

AtReconstruction/AtPatternModification/AtPatternModification.h
RealAurio: Changed in second commit after this one.

However, I had to drop the const on AtRawEvent and AtEvent. This is because, in the context of the Bragg curve fitter, I am using an AtPSA to get more resolution on the range values for the energy loss profile. The AtPSA do not take in const AtPads, so if I want to mark AtRawEvent as const, I would need to change AtPSA so that it takes const AtPads.

For the same reason I avoided creating a derived AtTrackBragg in this feature, I don't think this should be done here, since it would imply quite a big change that is no related to the purpose of this feature.

RealAurio: Changed in second commit after this one.

AtReconstruction/AtPatternModification/AtPatternModification.h
RealAurio: Changed in second commit after this one.

AtReconstruction/AtPatternModification/AtPatternModification.cxx
RealAurio: Changed in second commit after this one.

AtReconstruction/AtPatternModification/AtBraggCurveFinder.cxx
RealAurio: Changed in second commit after this one.

AtReconstruction/AtPatternModification/AtBraggCurveFinder.cxx
RealAurio: Changed in second commit after this one.

AtData/AtTrack.h
RealAurio: Same response as above comment.

AtData/AtTrack.h
RealAurio: Done in third commit after this one.

RealAurio: Done in third commit after this one.

AtEventDisplay/AtTabs/AtTabBraggCurve.cxx
RealAurio: Done in third commit after this one.

AtEventDisplay/AtTabs/AtTabBraggCurve.h
RealAurio: Done in third commit after this one.

AtReconstruction/AtPatternModification/AtPatternModification.h
RealAurio: Also, small detail. Instead of having a public method ModifyTrack, I implemented a protected method GetModifiedTrack that returns the modified track. This is because, since we are going to implement a derived AtTrack in the future, it makes sense to construct an object of the derived class internally and return it, since in general this method will take as input an AtTrack from the base class.

AtEventDisplay/AtTabs/AtTabBraggCurve.h
RealAurio: Done in third commit after this one.

However, I don't understand why this destructor should be virtual. Can you explain it to me?

AtReconstruction/AtPatternModificationTask.h
RealAurio: Defaults are given in the constructor.

AtReconstruction/AtPatternModificationTask.cxx
RealAurio: Changed in 4th commit after this one.

AtReconstruction/AtPatternModificationTask.h
RealAurio: Changed in 4th commit after this one.

AtReconstruction/AtPatternModificationTask.h
RealAurio: Changed in 4th commit after this one.

AtReconstruction/AtPatternModificationTask.h
RealAurio: Changed in 4th commit after this one.

AtReconstruction/AtPatternModificationTask.cxx
RealAurio: Changed in 4th commit after this one.

AtReconstruction/AtPatternModificationTask.cxx
RealAurio: Changed in 4th commit after this one.

I register it in any case. If both names are the same, then we would only have a problem if fIsPersistence is set to true in both the pattern recognition task and this task. In such a case, neither branch gets saved in the root file apparently. That case would not make sense though, so I don't worry about that.

AtReconstruction/AtPatternModificationTask.h
RealAurio: Changed in 4th commit after this one.

AtReconstruction/AtPulseAnalyzer/AtPSAHitPerTB.cxx
RealAurio: To be fair, on this one, the error log was already hardcoding a verbosity setting before I changed it. I just fixed a typo in this line. I will remove it though.

I'm pretty sure there are more cases in the framework where the logs are written like this, which explains why I started writing them like this, since I pretty much learned from this code.

From now on, if I ever notice something like this I will remove it and ship it in whatever feature I'm working at the moment.

AtReconstruction/AtReconstructionLinkDef.h
RealAurio: AtPatternModification and AtBraggCurveFinder can be - !, but AtPatternModificationTask needs to be +, or else it does not compile. Also, other tasks are saved as + here.

AtData/AtTrack.h
anthoak13: It still seems like these should live in their own derived class of AtTrack. But I don't remember if there was a reason we didn't do that?

AtReconstruction/AtPatternModification/AtPatternModification.cxx
```diff
@@ -0,0 +1,17 @@
+#include "AtPatternModification.h"
+
+void AtPatternModification::ModifyPatternEvent(AtPatternEvent *patternEvent, AtRawEvent *rawEvent, AtEvent *event)
```
anthoak13, line 3: Should either check for null pointers or we should pass AtPatternEvent as a reference rather than a pointer

AtReconstruction/AtPatternModification/AtPatternModification.cxx
```diff
@@ -0,0 +1,17 @@
...
+{
+   // Extract the candidate AtTracks.
+   std::vector<AtTrack> tracks = patternEvent->GetTrackCand();
+
+   // Create a new vector to store the modified AtTracks.
+   std::vector<AtTrack> modifiedTracks;
+
+   // Iterate over the original tracks and store the modified ones.
+   for (auto track : tracks)
+      modifiedTracks.push_back(GetModifiedTrack(&track, rawEvent, event));
```
anthoak13: Since we just have references anyway, GetModifiedTrack should probably take a reference rather than a pointer.

This use makes me think it should be a constant reference as well (we are returning a new track)

AtReconstruction/AtPatternModificationTask.cxx
anthoak13: Looks like no in-place modification

AtReconstruction/AtPatternModificationTask.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include <Rtypes.h>
+#include <TClonesArray.h>
+#include <TString.h>
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+class AtPatternModificationTask : public FairTask {
```
anthoak13, line 22: Should have a doxygen comment explaining this copies event to new branch and modifies each track.

AtReconstruction/AtPatternModificationTask.h
```diff
@@ -0,0 +1,50 @@
...
+private:
+   TString fInputBranchName;
+   TString fOutputBranchName;
+   TString fRawEventBranchName;
+   TString fEventBranchName;
```
anthoak13: Should have default values for input and output branch name, or those should required parts of the constructor. The class doesn't make sense as an object without them set. Best practices has an object functional an initialization, then use function calls to modify behavior

anthoak13 (changes requested):
I think this looks good with a few tiny comments. The big one is I don't remember the discussion on moving the brag-curve specific parts to their own derived class.

AtData/AtTrack.h
RealAurio: I'm not sure if you can see my comment above, but that's the reason why.

RealAurio: I didn't click on finish the review, that's why...

RealAurio (commented):
My comments that I forgot to submit...

AtReconstruction/AtPatternModification/AtPatternModification.cxx
RealAurio: Yeah, initially I just modified the AtTrack, but since we now want to keep the original AtPatternEvent, I can't just modify them.

Also, by returning a different AtTrack, this would allow this function to get a normal AtTrack and return a derived one, in the future when we implement derived AtTracks. But as mentioned earlier, this gave me problems, so for now I'll just work with normal AtTracks.

I will make it so that the input is a const reference to the original AtTrack.

AtReconstruction/AtPatternModificationTask.h
RealAurio: Will add.

AtReconstruction/AtPatternModificationTask.h
RealAurio: I have already commented on this, but forgot to press publish review.

I see in a lot of tasks, the default values are given in the constructor of the task, so that's why I put those there. In any case, I will also put the defaults here too even if it's redundant.

RealAurio: Done.

AtReconstruction/AtPatternModificationTask.h
RealAurio: Done.

AtReconstruction/AtPatternModification/AtPatternModification.cxx
RealAurio: Done.

AtReconstruction/AtPatternModification/AtPatternModification.cxx
RealAurio: Done.

AtData/AtTrack.h
anthoak13: That makes sense,!

AtReconstruction/AtPulseAnalyzer/AtPSAHitPerTB.cxx
anthoak13: Yeah there are a lot of sub-optimal things we should try to fix as we slowly update thing!

anthoak13 (approved):
I think this is ready to merge!

PR #259, opened 2026-03-03
Add Bethe Bloch energy loss model 

anthoak13:
Adds Bethe Bloch as an energy loss model to the code.

This was mostly done as a test of how Claude Code interacts with out codebase, but the results were very good and is probably worth including. Below you can see the results of this class compared to the expected output from wikipedia.

Screenshot 2026-03-03 135339

Screenshot 2026-03-03 135451

### DIFF
--- a/.claude/2026-03-03-131255-implement-beth-bloch.txt
+++ b/.claude/2026-03-03-131255-implement-beth-bloch.txt
@@ -1,0 +1,1257 @@
+
+╭─── Claude Code v2.1.63 ──────────────────────────────────────────────────────╮
+│                                                    │ Tips for getting        │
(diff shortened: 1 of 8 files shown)


anthoak13:
@Yassid @RealAurio I added the Bethe-Bloch mostly as a test, but I think it could be useful for the pion work?

RealAurio:
> Bethe

Hello @anthoak13, maybe @gxifrag will be interested in this too, she is the one working with the Kalman Filter right now.

Yassid:
@anthoak13  Yes this is a good idea. We are doing extensive tests of the KF performance and this is certainly critical.

PR #262, opened 2026-04-10
Rcnp analysis branch

anthoak13:


RealAurio:
@anthoak13 First of all, sorry for the spam.

I think it may be better to first merge the other feature branches this one is based upon first, so that it is more clear what features have been added in different commits, instead of one giant commit with way too many changes.

I would do the following pull requests first before attempting this one:
1. fitterRefactor: This has not been merger yet, but I think the pull request is still open, but it may need changes since there were changes in develop. This is the pull request #254 

2. experimentalBraggCurve: It had 1 last commit after it was merged, since I had to fix how the vertex finder is used. It  is a small change that only affects the AtPatternModification and the AtBraggCurveFinder classes, so it is easy to merge I think. I just created the pull request, which is #263 

3. driftDirectionInversion: It has some changes to the digitalization and PSA logic to allow for the case of the AT-TPC being inversed, such as the RCNP experiments and the S800 experiments. It probably needs reviewing still. Actually, I opened a pull request for this months ago, which is #258 

4. vetoBeamPads: This feature simply allows to pass the AtMap a list of beam pads, which will prevent the unpacker from reading them. Just opened the pull request #264 

5. IonChamberAnalysis: Some IC analysis code that we are using in RCNP analysis. It actually does not depend on anything from this branch, so it could also be merged. I just opened the pull request #265 

6. ENCourseMerger: This feature adds a merger class for the ENCourse daq. The pull request is #266 

7. AtMapCalibration: Finally this feature branch allows the user to pass calibration functions that may be used in AtPSA for example to give each AtHit a calibrated ELoss value. It is implemented but not really used yet, since people are still working on the analysis and calibrations. I created the pull request anyways, so that we can start iterating on it #267

PR #265, opened 2026-04-10
Added IC analysis classes

RealAurio:
This adds some IC analysis classes, since it was not so clear how to access it before.

With these changes, the IC analysis would be done using a task. This may need some more development though, but I will open the pull request anyways since this was merged into RCNPAnalysisBranch, but it does not actually depend on anything from it.


RealAurio:
I do not know exactly how this works, but this was coded by @pemi7269 . I assume she will get co-authoring in the merged commit?

